PR #10, opened 2023-10-17
Spack package implementation with Ascent rebuild pipeline

cameronrutherford:
Need to make sure that Ascent rebuild can push to GitHub, and then this is good to go! See Spack PR here: https://github.com/spack/spack/pull/40149


cameronrutherford:
LGTM @pelesh and @ryandanehy I think this is ready

cameronrutherford:
I still didn't update docs like I said I would, so perhaps we just make another issue so I get back to that... this is basically the same as ExaGO scripts so I would just be re-typing what's mostly in there with a few modifications

pelesh:
> I still didn't update docs like I said I would, so perhaps we just make another issue so I get back to that... this is basically the same as ExaGO scripts so I would just be re-typing what's mostly in there with a few modifications

More documentation would be helpful, especially for Spack related issues, but I agree, we need a first stab at this in the main development branch.

pelesh (approved):

cameronrutherford:
@ryandanehy feature or bug, the commit that's automated runs all testing again. We can change that by adding [ascent-test], but I would love to find a way to skip GitHub actions with different rules based on strings in commit messages, but am yet to bother.

PR #20, opened 2023-10-20
Make CUDA optional dependency

pelesh:
This PR reduces internal dependencies on CUDA SDK libraries. Classes `MatrixHandler` and `VectorHandler` are refactored to allow CPU-only or CUDA implementation of matrix and vector operations to be selected at runtime. With these changes:
- ReSolve can be built without CUDA
- Hardware backend (CPU or CUDA) is selected at runtime
- The code now builds on Linux and MacOS
- Fixed bug in FGMRES destructor

The build configuration allows for multiple hardware backends to coexist. In theory, we should be able to add HIP backend and build it at the same time as CUDA backend. If no GPU backend is built, a dummy device backend needs to be built instead (see `CpuMemory.hpp`).

Known issues not addressed in this PR:
-  `matvec` function implemented in `MatrixHandlerCuda` class corrupts matrix descriptor object.
- There are memory leaks in unit tests for CUDA implementations of matrix and vector handlers as CUDA workspace is not deleted.


resolve/vector/VectorHandler.cpp
```diff
@@ -14,17 +20,36 @@ namespace ReSolve {
...
    */
-  VectorHandler:: VectorHandler(LinAlgWorkspace* new_workspace)
+  VectorHandler::VectorHandler(LinAlgWorkspaceCUDA* new_workspace)
   {
-    workspace_ = new_workspace;
+    cudaImpl_ = new VectorHandlerCuda(new_workspace);
+    cpuImpl_  = new  VectorHandlerCpu();
+
+    isCudaEnabled_ = true;
+    isCpuEnabled_  = true;
```
kswirydo, line 50: Why is this true for Cpu?

tests/functionality/testKLU_Rf.cpp
```diff
@@ -30,7 +30,7 @@ int main(int argc, char *argv[])
   int error_sum = 0;
   int status = 0;
 
-  ReSolve::LinAlgWorkspaceCUDA* workspace_CUDA = new ReSolve::LinAlgWorkspaceCUDA;
+  ReSolve::LinAlgWorkspaceCUDA* workspace_CUDA = new ReSolve::LinAlgWorkspaceCUDA();
```
kswirydo, line 33: Not sure what changed here.

kswirydo (commented):
I thought about it for some time and there was probably an easier solution to make all lin alg operatios (in VectorHandler and MatrixHandler) run with various backends, but it would had been a CMake nightmare and not necessarily backward compatible. I think the stuff in this PR is pretty good. I had some minor comments. Also, why IndexPlusValue changed its name?

tests/functionality/testKLU_Rf.cpp
kswirydo: Actually I see it now once it is all in one line.

resolve/vector/VectorHandler.cpp
pelesh: If you are deciding at runtime (using conditional statements) which handler implementation you will use, then I believe you need both of them instantiated.

examples/r_KLU_KLU_standalone.cpp
```diff
@@ -95,15 +95,15 @@ int main(int argc, char *argv[])
...
-  vec_r->update(rhs, "cpu", "cuda");
+  vec_r->update(rhs, "cpu", "cpu");
 
-  matrix_handler->setValuesChanged(true);
+  matrix_handler->setValuesChanged(true, "cpu");
 
-  matrix_handler->matvec(A, vec_x, vec_r, &ONE, &MINUSONE, "csr", "cuda"); 
+  matrix_handler->matvec(A, vec_x, vec_r, &ONE, &MINUSONE, "csr", "cpu"); 
   real_type* test = vec_r->getData("cpu");
   (void) test; // TODO: Do we need `test` variable in this example?
```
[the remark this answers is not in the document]
pelesh: Fixed in [126b16a](https://github.com/ORNL/ReSolve/pull/20/commits/126b16aa0f1fe5a9ac11a4c62b232eb78da0c74b).

examples/r_KLU_KLU_standalone.cpp
```diff
@@ -95,15 +95,15 @@ int main(int argc, char *argv[])
...
```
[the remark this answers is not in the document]
pelesh: Fixed in [126b16a](https://github.com/ORNL/ReSolve/pull/20/commits/126b16aa0f1fe5a9ac11a4c62b232eb78da0c74b).

examples/r_KLU_KLU_standalone.cpp
```diff
@@ -95,15 +95,15 @@ int main(int argc, char *argv[])
...
 
-  printf("\t 2-Norm of the residual: %16.16e\n", sqrt(vector_handler->dot(vec_r, vec_r, "cuda")));
+  printf("\t 2-Norm of the residual: %16.16e\n", sqrt(vector_handler->dot(vec_r, vec_r, "cpu")));
```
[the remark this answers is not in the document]
pelesh: Fixed in [126b16a](https://github.com/ORNL/ReSolve/pull/20/commits/126b16aa0f1fe5a9ac11a4c62b232eb78da0c74b).

kswirydo (approved):

resolve/matrix/MatrixHandler.cpp
```diff
@@ -1,82 +1,98 @@
...
 #include <resolve/matrix/Csc.hpp>
 #include <resolve/matrix/Csr.hpp>
+#include <resolve/workspace/LinAlgWorkspace.hpp>
+#include <resolve/utilities/misc/IndexValuePair.hpp>
 #include "MatrixHandler.hpp"
+#include "MatrixHandlerCpu.hpp"
+
+#ifdef RESOLVE_USE_CUDA
+#include "MatrixHandlerCuda.hpp"
+#endif
```
cameronrutherford, line 15:
```suggestion
#include "MatrixHandlerCpu.hpp"

#ifdef RESOLVE_USE_CUDA
#include "MatrixHandlerCuda.hpp"
#else
#include "MatrixHandlerCpu.hpp"
#endif
```
I assume we simply always want the CPU backend, so this suggestion is probably wrong.

resolve/vector/CMakeLists.txt
```diff
@@ -1,32 +1,49 @@
...
-    # Build shared library ReSolve::vector
-    add_library(resolve_vector SHARED ${Vector_SRC})
-    target_link_libraries(resolve_vector PUBLIC resolve_backend_cpu)
+  target_link_libraries(resolve_vector PUBLIC resolve_backend_cuda)
 endif()
 
+# If no GPU is enabled link to dummy device backend
+if(NOT RESOLVE_USE_GPU)
+  target_link_libraries(resolve_vector PUBLIC resolve_backend_cpu)
+endif(NOT RESOLVE_USE_GPU)
```
cameronrutherford, line 44: Trying to better understand this. We either always want the cpu backend linked, or this is a dummy operation and does nothing, so we can remove it.

resolve/workspace/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+if(RESOLVE_USE_CUDA)
+  set(ReSolve_Workspace_SRC ${ReSolve_Workspace_SRC} ${ReSolve_Workspace_CUDASDK_SRC})
+endif()
+
+add_library(resolve_workspace SHARED ${ReSolve_Workspace_SRC})
+
+# If CUDA is enabled, link to ReSolve CUDA backend
+if(RESOLVE_USE_CUDA)
+  target_link_libraries(resolve_workspace PUBLIC resolve_backend_cuda)
+endif(RESOLVE_USE_CUDA)
```
cameronrutherford, line 35: We also don't link in the dummy backend here, so want to make sure I am on the same page.

Also maybe we need better docs, but hard to grasp conceptually what each component here is doing just staring at PR

resolve/workspace/LinAlgWorkspace.hpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include <resolve/workspace/LinAlgWorkspaceCpu.hpp>
+
```
cameronrutherford, line 4: Similarly suspicious here

tests/functionality/testKLU.cpp
```diff
@@ -1,6 +1,7 @@
 #include <string>
 #include <iostream>
 #include <iomanip>
+#include <cmath>
```
cameronrutherford, line 4:
```suggestion
#include <iostream>
#include <iomanip>
```

cameronrutherford: I also question whether <iomanip> should be included here and/or in other parts of resolve where it is seemingly unused.

cameronrutherford (approved):
With better docs will come enlightenment. LGTM

EDIT: We should make issues for:
- Future work
- Known bugs
- CI/CD additions

ryandanehy:
> @pelesh PNNL GitHub supports MacOS and Windows runners. Can we add that for ReSolve in GitHub actions?
> 
> Additionally, now ReSolve supports building on Linux w/ Spack in GitHub actions, so we should add a pipeline and testing for those builds without CUDA.

resolve/vector/CMakeLists.txt
pelesh: When accelerator device (CUDA or HIP) is not supported, one still needs to implement functions for device memory management. In this case, the "dummy" functions will either do nothing (e.g. in case of device synchronize call) or will report an error in case one tries to copy to unsupported device. There is probably a better way to go about this, but so far this is the best could come up with.

resolve/workspace/CMakeLists.txt
pelesh: We don't need dummy backend for CUDA-SDK and ROCm functions, because they are compiled with C++ compiler. So in this case, it was easier to find a clean solution.

pelesh:
> > @pelesh PNNL GitHub supports MacOS and Windows runners. Can we add that for ReSolve in GitHub actions?
> > Additionally, now ReSolve supports building on Linux w/ Spack in GitHub actions, so we should add a pipeline and testing for those builds without CUDA.
> 
> Will that mean we have to setup a mirror to a PNNL github given this under ORNL?

I think so.

resolve/workspace/LinAlgWorkspace.hpp
pelesh: It is different logic for workspaces and for memory utilities. Memory utilities are device specific (probably should have named that class `DeviceMemUtils`). On the other hand, the workspace should allocate both, host and device memory spaces, so the CPU workspace can coexist with other workspaces.

resolve/matrix/MatrixHandler.cpp
pelesh: This is correct. We can always build CPU workspace, so we want to have matrix handler built, as well. Otherwise, you would not be able to run computations on CPU if you enable e.g. CUDA backend.

resolve/vector/CMakeLists.txt
pelesh: I should perhaps add that host functions for memory management are always built. Having a "plug" for device memory management functions is a way to eliminate quite a few preprocessor directives in the code.

Another important point, if we are able to pass memory space parameter from user code all the way to mem utils, that would eliminate current solution, but would require far more refactoring.

tests/functionality/testKLU.cpp
pelesh: Including `cmath` is required for `sqrt` function in examples. I am not sure where else in the code we are including it.

cameronrutherford:
> > @pelesh PNNL GitHub supports MacOS and Windows runners. Can we add that for ReSolve in GitHub actions?
> > Additionally, now ReSolve supports building on Linux w/ Spack in GitHub actions, so we should add a pipeline and testing for those builds without CUDA.
> 
> Will that mean we have to setup a mirror to a PNNL github given this under ORNL?

Mirror from ORNL GitHub to PNNL GitLab should be sufficient without PNNL GitHub in between

Issue #28, opened 2023-10-26
HIP/ROCm 5.6 include directory info not propagated through CMake interface libraries


pelesh:
I am not sure if I am doing something wrong here, but it seems that include directory information for ROCm 5.6 does not propagate through CMake configuration the way _I think_ it should. Here is a simple example to reproduce the issue:
```cmake
find_package(hip REQUIRED)

add_library(resolve_hip INTERFACE)
target_link_libraries(resolve_hip INTERFACE hip::device)

get_target_property(hip_includes hip::device INTERFACE_INCLUDE_DIRECTORIES)
message(STATUS "HIP include directories: ${hip_includes}")

get_target_property(resolve_hip_includes resolve_hip INTERFACE_INCLUDE_DIRECTORIES)
message(STATUS "ReSolve HIP include directories: ${resolve_hip_includes}")
```
When this code is executed, I get something like this in the output. 
```
...
-- HIP include directories: /opt/rocm-5.6.0/include
-- ReSolve HIP include directories: resolve_hip_includes-NOTFOUND
...
```
Do I need to export `hip::device` target before interfacing it to `resolve_hip`?

I could use some help to understand this better.

cameronrutherford:
I feel like this is a feature not a bug, but maybe this is something that is CMake bug....

What I suggest would fix this:

```
find_package(hip REQUIRED)

add_library(resolve_hip INTERFACE)
target_link_libraries(resolve_hip INTERFACE hip::device)

get_target_property(hip_includes hip::device INTERFACE_INCLUDE_DIRECTORIES)
message(STATUS "HIP include directories: ${hip_includes}")

target_include_directories(resolve_hip INTERFACE ${hip_includes})
get_target_property(resolve_hip_includes resolve_hip INTERFACE_INCLUDE_DIRECTORIES)
message(STATUS "ReSolve HIP include directories: ${resolve_hip_includes}")
```

or, this might work?

```
find_package(hip REQUIRED)

add_library(resolve_hip INTERFACE)
target_link_libraries(resolve_hip PUBLIC hip::device)
```

pelesh:
The workaround currently used in the code partially addresses this issue:
```cmake
get_target_property(hip_includes hip::device INTERFACE_INCLUDE_DIRECTORIES)

target_include_directories(resolve_hip INTERFACE 
  $<BUILD_INTERFACE:${hip_includes}>)
```
I guess the root cause is in CMake's HIP support. Closing this issue for now.

PR #30, opened 2023-10-26
HIP backend

### LINKED ISSUE #12: Consider using enums instead of strings for user input
@rothpc says: I disagree strongly with the design decision to use strings to specify memory spaces.  The runtime cost of repeated string comparisons at run time will be high, and the error checking currently in the code (e.g., in Vector::getData) is insufficient (i.e., doesn't handle the situation where memspace is an unrecognized string).  I think it also keeps the code from being written such that the compiler can disallow nonsense calls (like the CPU call to allocate memory on device discussed above) as opposed to using runtime checks.

CC @kswirydo

### LINKED ISSUE #27: Add HIP/ROCm version of matvec function to MatrixHandler class
As an initial step to porting ReSolve to HIP we might want to start with HIP/ROCm implementation of matvec function. For starters, it would be sufficient to have it running within unit test framework. This work would also require implementing HIP-based linear algebra workspace.

Such development should branch off `hip-first-take-dev` branch. To build this branch on Frontier one needs to load modules PrgEnv-amd, amd/5.6.0, and cmake. We need our own build of SuiteSparse:
```shell
module use -a /lustre/orion/csc359/proj-shared/nkouk/spack-install/modules/linux-sles15-zen3
module load suite-sparse/5.13.0-clang-14.0.0-rocm5.2.0-mixed-ebugiv3
```
To build use:
```
CC=amdclang CXX=amdclang++ cmake -DRESOLVE_USE_HIP ../resolve
make
```
This assumes your build dir is a sibling of the ReSolve source dir.

Note that the `hip-first-take-dev` branch is under heavy development.

### LINKED ISSUE #29: 'delete' applied to a pointer that was allocated with 'new[]' in KLU examples/tests
The temporary storage arrays `x` and `rhs` used in examples and functionality tests are incorrectly deleted. With GCC, this triggers mysterious "unused variable" warnings.

CC @kswirydo

### LINKED ISSUE #37: Implement refactorization-based sparse direct linear solver for AMD GPUs
Leverage [ROCm libraries](https://rocsolver.readthedocs.io/en/latest/api/refact.html#rfsolver) to add direct sparse linear solvers that run on AMD GPUs. Follow similar approach used to implement CUDA-based direct solvers.

This is the main objective of the Frontier hackathon.

CC @maksud @rothpc @nkoukpaizan @cameronrutherford @ryandanehy @jaelynlitz

### LINKED ISSUE #45: Implement HIP API backed for ReSolve library
Create abstract interface to HIP API functions. Allow users to select between CUDA and HIP backends.

pelesh:
This is pre-hackathon PR with purpose to track multiple issues and code preparation for the event. So far, this includes:
* HIP backend to memory utilities
* CMake build for HIP (lots of patchwork)
* Fixes #29 and thus removes annoying warnings
* Resolves #27
* Resolves #12 
* Resolves #45 
* Resolves #37

CC @rothpc @cameronrutherford @jaelynlitz @ryandanehy @nkoukpaizan


pelesh:
This PR contains a workaround for #28.

resolve/hip/MemoryUtils.hip
```diff
@@ -0,0 +1,40 @@
+/**
```
cameronrutherford, line 1: Should this be `*.hpp` instead of `*.hip`?

pelesh:
Merged @kswirydo's implementation of matrix and vector kernels and rebased to `develop`. All CI pipelines pass, including the incline.

resolve/matrix/MatrixHandler.cpp
```diff
@@ -41,6 +44,7 @@ namespace ReSolve {
   {
     if (isCpuEnabled_)  delete cpuImpl_;
```
kswirydo, line 45: I think cpu is always enabled, no?

resolve/vector/VectorHandler.hpp
```diff
@@ -55,9 +57,11 @@ namespace ReSolve { //namespace vector {
     private:
       VectorHandlerImpl*  cpuImpl_{nullptr};
       VectorHandlerImpl* cudaImpl_{nullptr};
+      VectorHandlerImpl*  hipImpl_{nullptr};
 
       bool isCpuEnabled_{false};
```
kswirydo, line 62: I thought cpu is always enabled,no?

tests/unit/vector/VectorHandlerTests.hpp
```diff
@@ -241,14 +285,15 @@ namespace ReSolve {
         {
           bool status = true;
           if (memspace != "cpu") {
-            x->copyData(memspace, "cpu");
+            x->copyData(memory::DEVICE, memory::HOST);
```
kswirydo, line 288: wait how is this correct?

kswirydo (approved):

tests/unit/vector/VectorHandlerTests.hpp
pelesh: If `memspace != "cpu"` this means the memory is allocated either on CUDA or HIP device. So the copy in vector class should go from the device to the host.

resolve/hip/hipKernels.hip
```diff
@@ -0,0 +1,211 @@
...
+  int bsize = blockDim.x;
+
+  // assume T threads per thread block (and k reductions to be performed)
+  volatile __shared__ double s_tmp1[Tv5];
+
+  volatile __shared__ double s_tmp2[Tv5];
+  // map between thread index space and the problem index space
+  int j = blockIdx.x;
+  s_tmp1[t] = 0.0f;
+  s_tmp2[t] = 0.0f;
```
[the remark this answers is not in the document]
pelesh: Documented in #65, will be addressed separate from this PR.

resolve/hip/hipKernels.hip
```diff
@@ -0,0 +1,211 @@
...
+                                   double* result) {
+
+  // one thread per row, pass through rows
+  // and sum
+  // can be done through atomics
+  //\sum_{j=1}^m abs(a_{ij})
+
+  int idx = blockIdx.x*blockDim.x + threadIdx.x;
+  while (idx < n){
+    double sum = 0.0f;
```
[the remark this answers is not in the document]
pelesh: Documented in #65, will be addressed separate from this PR.

tests/unit/vector/VectorHandlerTests.hpp
kswirydo: ok

resolve/matrix/MatrixHandler.cpp
pelesh: Yes, that is correct. This conditional is redundant.

resolve/vector/VectorHandler.hpp
pelesh: The CPU handler should be indeed always enabled. However, it is not enabled until the vector handler constructor instantiates correct objects. I kept this flag because I think it might be helpful for debugging.

PR #41, opened 2023-10-30
Fix warnings in HIP branch.

pelesh:
Fix warnings due to unused variables and implicity type conversion between `size_t` and `index_type`.

### DIFF
--- a/resolve/workspace/LinAlgWorkspaceCpu.cpp
+++ b/resolve/workspace/LinAlgWorkspaceCpu.cpp
@@ -1,3 +1,4 @@
+#include <cstddef>
 #include "LinAlgWorkspaceCpu.hpp"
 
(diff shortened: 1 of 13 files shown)


kswirydo (approved):
I compiled it on Frontier. Tests run. Looked through the changes. Didnt spot anything weird.

resolve/workspace/LinAlgWorkspaceCpu.cpp
```diff
@@ -1,3 +1,4 @@
+#include <cstddef>
```
cameronrutherford, line 1: Is this header necessary?

pelesh: Yes, it contains definition of `size_t`.

PR #49, opened 2023-11-01
Fix segfault in examples

pelesh:
Delete all objects created in examples and avoid double delete. This is quick fix to `rocsolver-dev` branch.


kswirydo (approved):
Looks all good tome.

PR #50, opened 2023-11-01
Enable HIP support in Gramm-Schmidt and FGMRES

pelesh:
FGMRES solver and GS now build with HIP. This is still a hack. Example `klu_rocsolverrf_fgmres.exe` builds cleanly, but segfaults at execution. Needs some debugging.

This is a proposal how to get rid of hard-wired CUDA ID parameters.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -39,11 +39,15 @@ if(RESOLVE_USE_CUDA)
 
 endif(RESOLVE_USE_CUDA)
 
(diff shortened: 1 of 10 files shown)


pelesh:
With [c584383](https://github.com/ORNL/ReSolve/pull/50/commits/c584383fd929ed7505dedb80a88bf3d10f148418) I get the example to run without segfaulting, but IR diverges.

kswirydo (approved):
This all works

PR #56, opened 2023-11-02
A working alternative triangular solver (faster) for rocsolverrf

pelesh:
There is an option to choose between different triangular solvers from ROCm library. There is a significant difference in performance between those solvers depending on the problem at hand.


kswirydo (approved):
Ready to merge imho.

PR #59, opened 2023-11-02
Refactor CMake

### LINKED ISSUE #15: Use target commands to specify include directories in ReSolve
Currently ReSolve's root CMake file contains line setting:
```cmake
include_directories(${CMAKE_SOURCE_DIR})
```
This specifies source directory as the build include directory. This should not be necessary as each ReSolve target already specifies include directories (both for build and installation). However, without this option set, CMake does not add correct include lines to the make files it generates.

It is unclear why is this happening. This also could be a bug in CMake. This needs to be further investigated.

### LINKED ISSUE #46: Consider adding CMake macro to add ReSolve library
Consider adding a CMake macro similar to the onein [GridKit](https://github.com/ORNL/GridKit/blob/main/CMake/GridkitAddLibrary.cmake).

cameronrutherford:
- removes blanket `include_directories` everywhere (closes #15)

I will close remainder of #15 for AMD in Incline PR #53

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -71,8 +71,11 @@ set(CONSUMER_PATH ${CMAKE_INSTALL_PREFIX}/share/examples)
 install(PROGRAMS test.sh DESTINATION ${CONSUMER_PATH})
 
 # Select consumer app
+# TODO - have an outer loop that adds a unique consumer test for each backend supproted
 if(RESOLVE_USE_CUDA)
   set(RESOLVE_CONSUMER_APP "testKLU_Rf_FGMRES.cpp")
+elseif(RESOLVE_USE_HIP)
+  set(RESOLVE_CONSUMER_APP "testKLU_RocSolver.cpp")
 else()
   set(RESOLVE_CONSUMER_APP "testKLU.cpp")
 endif()
--- a/examples/resolve_consumer/CMakeLists.txt
+++ b/examples/resolve_consumer/CMakeLists.txt
@@ -11,7 +11,7 @@ find_package(ReSolve REQUIRED)
 
 # Build example with KLU factorization and KLU refactorization
 add_executable(consume.exe consumer.cpp)
-target_link_libraries(consume.exe PUBLIC ReSolve::ReSolve)
+target_link_libraries(consume.exe PRIVATE ReSolve::ReSolve)
 
 #------------------------------------------------------------------------------------
 # Testing of exported Resolve Configurations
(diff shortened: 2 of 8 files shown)


examples/CMakeLists.txt
```diff
@@ -71,8 +71,11 @@ set(CONSUMER_PATH ${CMAKE_INSTALL_PREFIX}/share/examples)
 install(PROGRAMS test.sh DESTINATION ${CONSUMER_PATH})
 
 # Select consumer app
+# TODO - have an outer loop that adds a unique consumer test for each backend supproted
```
cameronrutherford, line 74: @pelesh please create an issue that helps clarify this - if someone builds with CUDA and HIP, should we test linking against one? Both? Should we just leave it as is and remove this TODO comment?

cameronrutherford:
Ready to merge as change is atomic and tests are passing

examples/resolve_consumer/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ find_package(ReSolve REQUIRED)
 
 # Build example with KLU factorization and KLU refactorization
 add_executable(consume.exe consumer.cpp)
-target_link_libraries(consume.exe PRIVATE ReSolve::ReSolve)
+target_link_libraries(consume.exe PUBLIC ReSolve::ReSolve)
```
cameronrutherford, line 14:
```suggestion
target_link_libraries(consume.exe PRIVATE ReSolve::ReSolve)
```

examples/CMakeLists.txt
cameronrutherford: See #61

pelesh (changes requested):
When running `make test_install` the consumer app build fails with following error:
```
-- HIP include directories: /opt/rocm-5.6.0/include
-- Configuring done
CMake Error at /ccs/home/peles/local/resolve-frontier/share/resolve/cmake/ReSolveTargets.cmake:64 (set_target_properties):
  The link interface of target "ReSolve::resolve_hip" contains:

    roc::rocblas

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /ccs/home/peles/local/resolve-frontier/share/resolve/cmake/ReSolveConfig.cmake:5 (include)
  CMakeLists.txt:10 (find_package)
```
I think this needs just a little tweak to get it to work.

Otherwise changes look good!

PR #62, opened 2023-11-03
Fix warnings stemming from rocsolver class

pelesh:
Implementing ROCm-based solver introduced several warnings due to type conversion and method shadowing. This PR fixes those warnings and does some minor code cleanup.


kswirydo (approved):
1)It compiles with no warnings now
2) the tests are passing ok.

PR #73, opened 2023-11-10
Fixed include files

### LINKED ISSUE #72: Missing header file include in the installed include files
For the HIP version:
The include file `resolve/LinSolverDirectRocSolverRf.hpp` needs to add `#include <resolve/matrix/Csr.hpp>`.

For the CUDA version:
The include file `resolve/LinSolverDirectCuSolverGLU.hpp` needs to add `#include <resolve/workspace/LinAlgWorkspace.hpp>`

maksud:
Added the required include lines.

Fixes #72

### DIFF
--- a/resolve/LinSolverDirectCuSolverGLU.hpp
+++ b/resolve/LinSolverDirectCuSolverGLU.hpp
@@ -18,7 +18,7 @@ namespace ReSolve
   }
 
   // Forward declaration of ReSolve handlers workspace
-  class LinAlgWorkspace;
+  class LinAlgWorkspaceCUDA;
 
   class LinSolverDirectCuSolverGLU : public LinSolverDirect 
   {
--- a/resolve/LinSolverDirectCuSolverGLU.hpp
+++ b/resolve/LinSolverDirectCuSolverGLU.hpp
@@ -3,6 +3,7 @@
 #include "LinSolver.hpp"
 #include "cusolver_defs.hpp"
 #include <resolve/MemoryUtils.hpp>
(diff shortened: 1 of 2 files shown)


resolve/LinSolverDirectCuSolverGLU.hpp
```diff
@@ -3,6 +3,7 @@
 #include "LinSolver.hpp"
 #include "cusolver_defs.hpp"
 #include <resolve/MemoryUtils.hpp>
+#include <resolve/workspace/LinAlgWorkspace.hpp>
```
pelesh, line 7: This should not be included here. Use forward declaration instead. The source file should have specific includes.
```suggestion

```

pelesh: The bug here is on line 22. The forward declaration should be (I think):
```c++
  class LinAlgWorkspaceCUDA;
```

resolve/LinSolverDirectRocSolverRf.hpp
```diff
@@ -8,6 +8,7 @@
 #include <rocblas/rocblas.h>
 #include <rocsolver/rocsolver.h>
 #include <hip/hip_runtime.h>
+#include <resolve/matrix/Csr.hpp>
```
pelesh, line 12: Use forward declaration here. If necessary, add this include to the source file.
```suggestion

```

pelesh (changes requested):
Please use forward declarations in header files and include headers to the source files as needed. We need to keep build dependencies minimal in order to maintain code portability.

pelesh (approved):
lgtm

PR #77, opened 2023-11-22
Added example with redoing the factorization through KLU

kswirydo:
What is new is this PR:

1. Fixed memory leaks in KLU - now if a new factorization is performed, I delete the existing Symbolic and Numeric objects first. I also free L_ and U_ so that if matrix (or its nnz pattern) changed, the factors would be extracted again. I set the variable ` factors_extracted_` accordingly.

2. Added example where based on a simple norm criteria, I redo the KLU numeric and symbolic and set up refactorization if this norm is too large. To test, use:
`./runResolve build/examples/klu_rocsolverrf_check_redo.exe`


pelesh (approved):
I understand the idea of this, but it seems with all (de)allocations, repeating numerical factorization on CPU will be extremely expensive.

PR #102, opened 2023-12-06
Matrix norm

kswirydo:
Matrix inf norm commonly appears in error evaluation formulae (i.e. NSR, etc). I added Cuda, CPU and HIP versions. It is not a part of standard BLAS.

Matrix Inf norm is max of the sum of rows (with abs value).
This is done as 3 kernels:
1) Sum rows, store in `d_r` (buffer).
2) Reduce `d_r` to another buffer (blockwise).
3) Second pass of reduction - reduce to a variable.
It is pretty straightforward to do on CPU.

There is a `cusolverSp` function that does 2+3 (which I am using). The rest I implemented myself. 
Should be easy to add unit test (?) Can add before merging.

### DIFF
--- a/resolve/LinSolverDirectCuSolverRf.cpp
+++ b/resolve/LinSolverDirectCuSolverRf.cpp
@@ -7,6 +7,7 @@ namespace ReSolve
   LinSolverDirectCuSolverRf::LinSolverDirectCuSolverRf()
   {
     cusolverRfCreate(&handle_cusolverrf_);
+    setup_completed_ = false;
   }
 
   LinSolverDirectCuSolverRf::~LinSolverDirectCuSolverRf()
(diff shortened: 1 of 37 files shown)


resolve/matrix/MatrixHandlerHip.cpp
```diff
@@ -103,9 +104,39 @@ namespace ReSolve {
...
   }
 
-  int MatrixHandlerHip::Matrix1Norm(matrix::Sparse* /* A */, real_type* /* norm */)
+  int MatrixHandlerHip::MatrixInfNorm(matrix::Sparse* A, real_type* norm)
   {
-    return -1;
+    // we assume A is in CSR format
+    real_type* d_r = workspace_->getDr();
+    printf("d_r size %d \n", workspace_->getDrSize());
+    if (workspace_->getDrSize() != A->getNumRows()) {
```
pelesh, line 112: This print statement violates xSDK requirements.

resolve/LinSolverDirectCuSolverRf.cpp
```diff
@@ -25,11 +26,28 @@ namespace ReSolve
...
+      mem_.allocateArrayOnDevice(&d_P_, n);
+    } 
+
+    if (d_Q_ == nullptr){
+      mem_.allocateArrayOnDevice(&d_Q_, n);
+    }
+
+    if (d_T_ != nullptr){
+      mem_.deleteOnDevice(d_T_);
+    }
```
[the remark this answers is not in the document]
pelesh: After looking closer at the code, my suggestion would not work.

resolve/matrix/MatrixHandlerHip.cpp
kswirydo: It was placed here for debugging/testing only.

resolve/vector/VectorHandler.cpp
```diff
@@ -129,6 +129,30 @@ namespace ReSolve {
     }
   }
 
+  /** 
+   * @brief compute infinity norm of a vector (i.e., find an entry with largest absolute value)
+   * 
+   * @param[in] The vector
+   * @param memspace[in] string containg memspace (cpu or cuda or hip)
+   *
```
pelesh, line 137:
```suggestion
   * @param[in] The vector
   * @param[in] memspace string containg memspace (cpu or cuda or hip)
   *
```

pelesh (approved):
Looks good!

PR #104, opened 2023-12-07
Suggested changes for PR102

pelesh:
Some simple changes/suggestions for PR #102.


kswirydo:
I cleaned everything up according to @pelesh  suggestions. Added vector inf norm, unit tests, etc.

PR #110, opened 2023-12-12
Refactor handlers

### LINKED ISSUE #99: getRandSketchingMethod should be called setRandSketchingMethod
I think this should return the sketching method, not take it as input, no? I.e. this function should be called `set`

_Originally posted by @kswirydo in https://github.com/ORNL/ReSolve/pull/86#discussion_r1414762112_

pelesh:
This PR features refactored handler classes. The changes are as follows:
- Memory space of the handler is set upon instantiation.
- Host handler accompanies each device handler.
- Selection between device and host handler is done in the same way as for matrix/vector classes.
- All examples updated.
- All tests updated and passing.
- Closes #99.

### DIFF
--- a/resolve/vector/VectorHandlerCpu.cpp
+++ b/resolve/vector/VectorHandlerCpu.cpp
@@ -119,7 +119,7 @@ namespace ReSolve {
    * @brief gemv computes matrix-vector product where both matrix and vectors are dense.
    *        i.e., x = beta*x +  alpha*V*y
    *
-   * @param[in] Transpose - yes (T) or no (N)
+   * @param[in] Transpose - transposed = 'T' or not 'N'
    * @param[in] n Number of rows in (non-transposed) matrix
    * @param[in] k Number of columns in (non-transposed)   
    * @param[in] alpha Constant real number
@@ -131,7 +131,7 @@ namespace ReSolve {
    * @pre   V is stored colum-wise, _n_ > 0, _k_ > 0
    * 
    */  
-  void VectorHandlerCpu::gemv(std::string /* transpose */,
+  void VectorHandlerCpu::gemv(char /* transpose */,
                               index_type /* n */,
                               index_type /* k */,
                               const real_type* /* alpha */,
--- a/resolve/vector/VectorHandlerCpu.hpp
+++ b/resolve/vector/VectorHandlerCpu.hpp
@@ -1,5 +1,4 @@
 #pragma once
-#include <string>
 
 namespace ReSolve
 { 
@@ -45,7 +44,7 @@ namespace ReSolve { //namespace vector {
        * if `transpose = T` (yes), `x = beta*x + alpha*V^T*y`,
        * where `x` is `[k x 1]`, `V` is `[n x k]` and `y` is `[n x 1]`.
        */ 
-      virtual void gemv(std::string transpose,
+      virtual void gemv(char transpose,
                         index_type n,
                         index_type k,
                         const real_type* alpha,
--- a/resolve/vector/VectorHandlerCuda.cpp
+++ b/resolve/vector/VectorHandlerCuda.cpp
@@ -140,7 +140,7 @@ namespace ReSolve {
    * @pre   V is stored colum-wise, _n_ > 0, _k_ > 0
    * 
    */  
-  void VectorHandlerCuda::gemv(std::string transpose,
+  void VectorHandlerCuda::gemv(char transpose,
                                index_type n,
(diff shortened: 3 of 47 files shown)


resolve/vector/VectorHandlerCpu.cpp
```diff
@@ -131,7 +131,7 @@ namespace ReSolve {
    * @pre   V is stored colum-wise, _n_ > 0, _k_ > 0
    * 
    */  
-  void VectorHandlerCpu::gemv(std::string /* transpose */,
+  void VectorHandlerCpu::gemv(char /* transpose */,
```
kswirydo, line 134: Why?

resolve/vector/VectorHandlerCpu.hpp
```diff
@@ -45,7 +44,7 @@ namespace ReSolve { //namespace vector {
        * if `transpose = T` (yes), `x = beta*x + alpha*V^T*y`,
        * where `x` is `[k x 1]`, `V` is `[n x k]` and `y` is `[n x 1]`.
        */ 
-      virtual void gemv(std::string transpose,
+      virtual void gemv(char transpose,
```
kswirydo, line 47: Same here

resolve/vector/VectorHandlerCuda.cpp
```diff
@@ -199,8 +198,7 @@ namespace ReSolve {
     if (k < 200) {
```
kswirydo, line 198: this constant would need to become a parameter at some point (but this is a future issue)

resolve/vector/VectorHandlerCuda.cpp
```diff
@@ -239,8 +237,7 @@ namespace ReSolve {
...
```
kswirydo, line 237: and same here.

resolve/vector/VectorHandlerCuda.hpp
```diff
@@ -45,7 +45,7 @@ namespace ReSolve { //namespace vector {
        * if `transpose = T` (yes), `x = beta*x + alpha*V^T*y`,
        * where `x` is `[k x 1]`, `V` is `[n x k]` and `y` is `[n x 1]`.
        */ 
-      virtual void gemv(std::string transpose,
+      virtual void gemv(char transpose,
```
kswirydo, line 48: why? I mean you should not compare chars like we compare strings,with `==`!!!

tests/unit/matrix/MatrixHandlerTests.hpp
```diff
@@ -34,12 +34,17 @@ class MatrixHandlerTests : TestBase
   TestOutcome matrixInfNorm(index_type N)
   {
     TestStatus status;
+    ReSolve::memory::MemorySpace ms;
```
kswirydo, line 37: what is the difference between `ms` and `memspace`??

tests/unit/vector/VectorHandlerTests.hpp
```diff
@@ -92,7 +92,7 @@ namespace ReSolve {
 
           real_type alpha = 0.5;
           //the result is a vector with y[i] = 2.5;          
-          handler->axpy(&alpha, x, y, memspace_);
+          handler->axpy(&alpha, x, y, ms);
```
kswirydo, line 95: again, what is the function of `ms` and `memspace`??

kswirydo (approved):

resolve/vector/VectorHandlerCpu.cpp
pelesh: It is cheaper to compare `char` than `std::string`. Plus you can use `char` in `switch` statements.

tests/unit/matrix/MatrixHandlerTests.hpp
pelesh: `memspace` is string and `ms` is `enum`. I will eventually get rid of all strings.

tests/unit/vector/VectorHandlerTests.hpp
pelesh: Ditto.

resolve/vector/VectorHandlerCuda.cpp
pelesh: See #112.

resolve/vector/VectorHandlerCuda.cpp
pelesh: To be tracked in #112.

PR #111, opened 2023-12-12
SDK compliance of adding api version

ryandanehy:
Add version.cpp function that grabs the version of resolve in order to satisfy SDK compliance. In reference to #2


ryandanehy:
I included a test within make test that runs the function so the pipelines should be doing tests of it.

resolve/utilities/version.hpp
```diff
@@ -0,0 +1,18 @@
+#ifndef RESOLVE_VERSION_H
+#define RESOLVE_VERSION_H
+
+
+/**
+ * Sets major, minor, and patch versions for current ReSolve build. The user is
+ * responsible for free'ing this memory.
+ */
+int ResolveVersionGetVersion(int *, int *, int *);
```
pelesh, line 9: Some suggested style formatting and comment corrections:
```suggestion
/**
 * Sets major, minor, and patch versions for current ReSolve build. The user is
 * responsible for ensuring `int`s are allocated.
 */
int ResolveVersionGetVersion(int*, int*, int*);
```

resolve/utilities/version.hpp
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Sets string with build version for current ExaGO build in format
+ * "major.minor.patch". The user is responsible for free'ing this memory.
+ */
+int ReSolveVersionGetVersionStr(std::string &);
+
+
+#endif
```
pelesh, line 18: End file with newline.
```suggestion
#endif

```

tests/functionality/testversion.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+
+int main(int argc, char *argv[])
+{
+  std::string versionstr;
+  ReSolveVersionGetVersionStr(versionstr);
+  std::cout<<"ReSolveVersionGetVersionStr Test: "<<versionstr<<std::endl<<std::endl;;
+  
+  return 0;
+}
```
pelesh, line 19: I would make this test verify the version, as well.
```suggestion
  std::string versionstr;
  std::string answer("1.0.0-alpha1");
  ReSolveVersionGetVersionStr(versionstr);
  std::cout<<"ReSolveVersionGetVersionStr Test: "<<versionstr<<std::endl<<std::endl;
  
  if (versionstr != answer) {
    std::cout << "Incorrect ReSolve version returned!\n";
    return 1;
  }
  return 0;
}

```
This will force future developers to check if the correct version is set ahead of the release.

pelesh (commented):
Since we are preparing alpha release, how will function `int ResolveVersionGetVersion(int *, int *, int *);` handle something like `1.0.0-alpha1`? Perhaps we could add version [label](https://github.com/LLNL/sundials/blob/v6.6.2/CMakeLists.txt#L52) to the versioning?

ryandanehy:
> Since we are preparing alpha release, how will function `int ResolveVersionGetVersion(int *, int *, int *);` handle something like `1.0.0-alpha1`? Perhaps we could add version [label](https://github.com/LLNL/sundials/blob/v6.6.2/CMakeLists.txt#L52) to the versioning?

This function was designed to make our lives potentially easier in future releases. I imagined a scenario in which we would want to change the cmake based on whether the version was less than or greater than a particularly major or minor release. In this situation the alpha part doesn't really matter in my opinion because any major change that would utlize this function would result in a change in major, minor or patch number right? And we always have the ReSolveVersionGetVersionstr function to get the whole version.

ryandanehy:
How are we feeling about this pr?

pelesh:
> How are we feeling about this pr?

I just added a couple of touches and managed to break the build :). I think if tests pass, we can merge this.

pelesh (approved):

PR #113, opened 2023-12-14
CPU vector handler functions

kswirydo:
Added:
- gemv
- mass axpy
- mass dot product with 2 vectors
- tests in cpu testing suite

Fixed:
- subtle bug in setToConst for multivectors

### DIFF
--- a/resolve/vector/Vector.cpp
+++ b/resolve/vector/Vector.cpp
@@ -295,7 +295,7 @@ namespace ReSolve { namespace vector {
           h_data_ = new real_type[n_ * k_]; 
           owns_cpu_data_ = true;
         }
-        for (int i = j * n_current_; i < (j + 1 ) * n_current_ * k_; ++i){
+        for (int i = j * n_current_; i < (j + 1 ) * n_current_ ; ++i){
           h_data_[i] = C;
         }
         break;
--- a/resolve/vector/Vector.cpp
+++ b/resolve/vector/Vector.cpp
@@ -295,7 +295,7 @@ namespace ReSolve { namespace vector {
           h_data_ = new real_type[n_ * k_]; 
           owns_cpu_data_ = true;
         }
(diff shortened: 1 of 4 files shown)


resolve/vector/Vector.cpp
```diff
@@ -295,7 +295,7 @@ namespace ReSolve { namespace vector {
           h_data_ = new real_type[n_ * k_]; 
           owns_cpu_data_ = true;
         }
-        for (int i = j * n_current_; i < (j + 1 ) * n_current_ * k_; ++i){
+        for (int i = j * n_current_; i < (j + 1 ) * n_current_ ; ++i){
           h_data_[i] = C;
```
pelesh, line 299: Coding style: 
```suggestion
        for (int i = j * n_current_; i < (j + 1 ) * n_current_ ; ++i) {
          h_data_[i] = C;
```

resolve/vector/VectorHandlerCpu.cpp
```diff
@@ -154,9 +183,27 @@ namespace ReSolve {
...
+    real_type* y_data = y->getData(memory::HOST);
+    real_type* x_data = x->getData(memory::HOST);
+    index_type i, j;
+    real_type sum;
+    
+    for (i = 0; i < size; ++i) {
+      sum = 0.0;
+      for (j = 0; j < k; ++j) {
+     // if(i == 2)  printf("size x %d size alpha %d multiplying x[%d] = %16.16e by alpha[%d] = %16.16e \n",size *k, k, j * size + i, x_data[j * size + i], j, alpha_data[j] );
+        sum += x_data[j * size + i] * alpha_data[j];
```
pelesh, line 203: Just get rid of the comment.
```suggestion
      for (j = 0; j < k; ++j) {
        sum += x_data[j * size + i] * alpha_data[j];
```

pelesh (approved):
Should we have all these functions return `int` instead of returning `void`?

In any case, good to merge with some minor fixes.

PR #116, opened 2023-12-15
Spack cpu build

### LINKED ISSUE #55: Spack CPU GitHub Action Build
Similar to ExaGO, we should support a minimal Spack based CPU build once we have a public spack package.

### LINKED ISSUE #61: FIx Consumer Example to build against all supported backends
From #59, only one potential backend is build in the consumer example. We should build all supported backends, as well as running tests for all fo them.

### LINKED ISSUE #90: CMake has in built list operands - use them in CmakeLists.
Not necessary to change, but CMake has in built list operands that are a little easier to read - https://cmake.org/cmake/help/latest/command/list.html

_Originally posted by @cameronrutherford in https://github.com/ORNL/ReSolve/pull/83#discussion_r1409386165_

### LINKED ISSUE #91: Install only needed headers
CUDA / ROCm specific header files should only be installed conditionally to lists like this

_Originally posted by @cameronrutherford in https://github.com/ORNL/ReSolve/pull/83#discussion_r1409387244_

### LINKED ISSUE #115: CPU Build Requires an Additional Include statement
When building from the cpu only cmake preset depending on compiler the following error occurs. 
Screenshot 2023-12-14 at 12 23 37 PM

Solution is to add 

`#include <cmath>` 

to SystemSolver.cpp

cameronrutherford:
Closes #55 and other linked issues
Closes #115
Closes #61
Closes #91 
Closes #90

### DIFF
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -1,0 +1,137 @@
+# https://spack.readthedocs.io/en/latest/binary_caches.html#spack-build-cache-for-github-actions
+name: Spack Ubunutu x86_64 Buildcache
+
+env:
+  SPACK_COLOR: always
+  REGISTRY: ghcr.io/ornl
+  # Our repo name contains upper case characters, so we can't use ${{ github.repository }}
+  IMAGE_NAME: resolve
+  USERNAME: resolve-bot
+  BASE_VERSION: ubuntu-22.04-fortran
+
+# Until we remove the need to clone submodules to build, this should on be in PRs
+on: [pull_request]
+
+jobs:
+  base_image_build:
+    runs-on: ubuntu-22.04
+    permissions:
+      packages: write
+      contents: read
+
+    name: Build Custom Base Image
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          # Once we move submodule deps into spack, we can do some more builds
+          # Also need to change build script to use spack from base image
+          submodules: true
+
+      # Need to build custom base image with gfortran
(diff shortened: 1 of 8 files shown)


.github/workflows/spack_cpu_build.yaml
```diff
@@ -0,0 +1,137 @@
...
+        run: spack -e . mirror set --oci-username ${{ env.USERNAME }} --oci-password "${{ secrets.GITHUB_TOKEN }}" local-buildcache
+      
+      - name: Trust keys
+        run: spack -e . buildcache keys --install --trust
+      
+      - name: Find external packages
+        run: spack -e . external find --all
+      
+      - name: Spack develop exago
+        run: spack -e . develop --path=$(pwd) resolve@develop
```
pelesh, line 125:
```suggestion
      - name: Spack develop ReSolve
        run: spack -e . develop --path=$(pwd) resolve@develop
```

.github/workflows/spack_cpu_build.yaml
```diff
@@ -0,0 +1,137 @@
...
+    
+    strategy:
+      matrix:
+        # Minimal Build(s) - GHCR mirror speeds these up a lot!
+        spack_spec:
+          - resolve@develop+klu
+          - resolve@develop~klu 
+
+    name: Build ExaGO with Spack
+    steps:
```
pelesh, line 88:
```suggestion
    name: Build ReSolve with Spack
    steps:
```

pelesh (approved):
Looks good and is ready to merge.

There is a couple of references to ExaGO that need to be removed.

PR #118, opened 2023-12-15
Add ctest to spack build pipelines.

cameronrutherford:
Adds a ctest invocation to the CPU spack builds (same fix can be extended to ExaGO + HiOp if functioning well...)

### DIFF
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -131,7 +131,7 @@ jobs:
         run: spack -e . install --no-check-signature
       
       - name: Test Installation
-        run: . buildsystem/spack/spack/share/spack/setup-env.sh && spack -e . load cmake && cd `ls | grep spack-build | grep -v txt` && ctest -VV
+        run: eval `spack load --sh cmake` && cd `ls | grep spack-build | grep -v txt` && ctest -VV
       
       # Push with force to override existing binaries...
       - name: Push to binaries to buildcache
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -128,8 +128,11 @@ jobs:
       - name: Concretize
         run: spack -e . concretize
 
-      - name: Install
-        run: spack -e . install --keep-stage --no-check-signature --no-cache
+      - name: Install dependencies
+        run: spack -e . install --no-check-signature --only dependencies
+
+      - name: Install package
+        run: spack -e . install --keep-stage --no-check-signature --no-cache --fresh
 
       - name: Test Installation
         run: cd $(spack -e . location --build-dir resolve@develop) && ctest -VV
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -129,7 +129,7 @@ jobs:
         run: spack -e . concretize
 
       - name: Install
-        run: spack -e . install --keep-stage --no-check-signature
+        run: spack -e . install --keep-stage --no-check-signature --no-cache
 
       - name: Test Installation
         run: cd $(spack -e . location --build-dir resolve@develop) && ctest -VV
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -131,7 +131,7 @@ jobs:
         run: spack -e . install --no-check-signature
       
       - name: Test Installation
-        run: eval `spack load --sh cmake` && cd `ls | grep spack-build | grep -v txt` && ctest -VV
+        run: cd `ls | grep spack-build | grep -v txt` && ctest -VV
       
       # Push with force to override existing binaries...
       - name: Push to binaries to buildcache
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -132,7 +132,7 @@ jobs:
         run: spack -e . install --no-check-signature
 
       - name: Test Installation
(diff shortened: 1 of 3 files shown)


cameronrutherford:
For some reason no tests are being found... Need to debug + diagnose further, as a similar workflow works just fine locally

superwhiskers:
i suggest we enable address sanitizer (and consider adding `ubsan` and `isan` to the mix) with `-DRESOLVE_USE_ASAN=yes` in the flags to cmake if we're going to run tests this way, and if not immediately, we should make this a priority, as it'd keep pull requests from introducing obvious memory safety issues (and if we add `ubsan`/`isan`, integer overflow and undefined behavior)

this, of course, breaks a lot of tests immediately due to the amount of memory leaks present, but i'm sure the current ones aren't too hard to fix

pelesh:
> For some reason no tests are being found... Need to debug + diagnose further, as a similar workflow works just fine locally

@cameronrutherford, do we know why tests are not run?

cameronrutherford:
> i suggest we enable address sanitizer (and consider adding `ubsan` and `isan` to the mix) with `-DRESOLVE_USE_ASAN=yes` in the flags to cmake if we're going to run tests this way, and if not immediately, we should make this a priority, as it'd keep pull requests from introducing obvious memory safety issues (and if we add `ubsan`/`isan`, integer overflow and undefined behavior)
> 
> this, of course, breaks a lot of tests immediately due to the amount of memory leaks present, but i'm sure the current ones aren't too hard to fix

This is a great idea and should be captured in another issue. Would just involve adding a flag to CMake through spack, and then also adding that to the `CMakePresets.json` for the relevant build configurations.

cameronrutherford:
> > For some reason no tests are being found... Need to debug + diagnose further, as a similar workflow works just fine locally
> 
> @cameronrutherford, do we know why tests are not run?

Part of the difficulty is that the GitHub actions runners are a VM, and not a strict container. Any debugging locally doesn't 1-1 map to the CI environment.

We could do the following:
- Rent an Azure VM to truly reproduce (worst idea)
- Add testing support to the spack package directly (see https://github.com/spack/spack/pull/44765 for HiOp example. Would just have to figure out a way to call `ctest` directly...)
- Keep running CI and debugging scripts... Maybe I just need to be POSIX compliant or there's an obvious way to fix this

cameronrutherford:
I force pushed to re-run CI and will try a few things, but I don't know if I will get anywhere, and the `spack test` support #152  seems like the more sensible approach perhaps? Maybe it would also be easier to support some easier way from running tests out of the installation directory, instead of the build directory like I am trying to do.

cameronrutherford:
> i suggest we enable address sanitizer (and consider adding `ubsan` and `isan` to the mix) with `-DRESOLVE_USE_ASAN=yes` in the flags to cmake if we're going to run tests this way, and if not immediately, we should make this a priority, as it'd keep pull requests from introducing obvious memory safety issues (and if we add `ubsan`/`isan`, integer overflow and undefined behavior)
> 
> this, of course, breaks a lot of tests immediately due to the amount of memory leaks present, but i'm sure the current ones aren't too hard to fix

We can easily add that to the tests run on PNNL / ORNL systems that have CI (ExCL will work, and

pelesh:
> I force pushed to re-run CI and will try a few things, but I don't know if I will get anywhere, and the `spack test` support #152 seems like the more sensible approach perhaps? Maybe it would also be easier to support some easier way from running tests out of the installation directory, instead of the build directory like I am trying to do.

Shouldn't we have both -- test build and test installation? We have separate `make test` and `make test_install` commands.

pelesh:
The issue I see here is that `ctest` is not run:
```
Test project /home/runner
Constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
No tests were found!!!
```

.github/workflows/spack_cpu_build.yaml
```diff
@@ -111,25 +111,29 @@ jobs:
...
       - name: Concretize
         run: spack -e . concretize
-  
+
       - name: Install
         run: spack -e . install --no-check-signature
-      
+
+      - name: Test Installation
+        run: cd $(spack -e . location --build-dir resolve@develop) && ctest -VV
```
cameronrutherford, line 135: This seems clean and worked locally...

buildsystem/spack/spack
cameronrutherford, line 1: This is the checkout that I have in the Spack PR. We could add builds with and without lusol once testing is working quite easily

cameronrutherford:
This is rebased on #168 and builds are running and so is ctest in CI after the spack builds!! ◡̈

cameronrutherford (commented):
I undid all the applied linting to the CMake, so I think this is ready to merge @pelesh

.github/workflows/spack_cpu_build.yaml
```diff
@@ -111,25 +112,35 @@ jobs:
...
-  
-      - name: Install
-        run: spack -e . install --no-check-signature
-      
+
+      - name: Install dependencies
+        run: spack -e . install --no-check-signature --only dependencies
+
+      - name: Install package
+        run: spack -e . install --keep-stage --no-check-signature --no-cache --fresh
```
cameronrutherford, line 136:
```suggestion
        run: spack -e . install --keep-stage --no-cache --fresh
```

I think one of `--no-cache` and `--fresh` aren't necessary, but we definitely don't need to check signatures at this stage as there shouldn't be anything to download...

pelesh (approved):
All tests run correctly and pass. Great job!

PR #125, opened 2023-12-17
Use memspace as parameter consistently in FGMRES classes.

pelesh:
FGMRES classes are now device agnostic. All is left to do is to make minor tweaks to Gram-Schmidt class, so we can run iterative solvers without GPUs.


kswirydo (approved):
All files look correct. I guess we will need an example to check if it really works....

Issue #63, opened 2023-11-03
Add platform agnostic profiling support`


cameronrutherford:
roctx support was added during the hackathon by @PaulMullowney for AMD platforms, but we should abstract this to work for any backend supported similar to how ExaGO does.

This should be reflected as well in the spack package, and hopefully by the time we finish this, `roctracer-dev`'s spack package works at the CMake level

cameronrutherford:
This is the extension of #68

cameronrutherford:
https://github.com/ROCm/rocprofiler/issues/131 - according to this, ROCTx now works with Spack just fine. This isn't blocked as far as I can tell

pelesh:
CC @superwhiskers

PR #134, opened 2024-01-22
Add tracer annotations to HIP code in ReSolve

### LINKED ISSUE #68: Add roctx support
https://github.com/ginkgo-project/ginkgo/blob/develop/cmake/Modules/FindROCTX.cmake

While in Frontier hackathon we couldn't get roctx to work on Incline, with the above find module suggested in issue https://github.com/ROCm-Developer-Tools/roctracer/issues/56 or some variation we should be able to get this to work.

I don't know if we have NVTX or plan on it, so maybe we re-target this issue towards supporting both profiling backends.

If we do this, we could think about supporting automatic benchmarking

pelesh:
Add some generic support for building ReSolve with profiling annotations and implement ROC Tracer annotations for HIP-based code in ReSolve. This is based on `no-splitlu-exp-roctx` branch by @PaulMullowney.

This will allow us to add in a straightforward manner annotations for CUDA (and other) profilers in the future.

Note that ROC Tracer does not provide CMake targets (see [roctracer issue 56](https://github.com/ROCm/roctracer/issues/56)), so the CMake configuration for it looks like a hack.

Closes #68.


kswirydo (approved):

resolve/CMakeLists.txt
```diff
@@ -135,10 +135,22 @@ target_include_directories(ReSolve INTERFACE
...
 target_link_libraries(ReSolve PUBLIC ${ReSolve_Targets_List})
 target_link_libraries(ReSolve PRIVATE resolve_version)
 
+if(RESOLVE_USE_PROFILING)
+  if(RESOLVE_USE_HIP)
+    # Roctracer does not provide CMake target, so we use this hack here.
+    # The assumption is roctracer lib and headers are installed at the same
+    # place as the rest of ROCm.
+    target_link_libraries(ReSolve PUBLIC "-lroctracer64 -lroctx64")
+  elseif(RESOLVE_USE_CUDA)
```
nkoukpaizan, line 147: I'd suggest throwing a warning or error here to notify the user that profiling is currently only support with roctx.

pelesh: I believe [918a794](https://github.com/ORNL/ReSolve/pull/134/commits/918a794cb541d5c1927403bdfe6b8a01413553a7) addresses this comment.

cameronrutherford: cc #63

resolve/CMakeLists.txt
```diff
@@ -135,10 +135,26 @@ target_include_directories(ReSolve INTERFACE
...
+    # Nothing to do for CUDA profiling for now.
+    message(NOTICE "Profiling support enabled, but Re::Solve does not create tracer annotations for CUDA.")
+    message(NOTICE "This profiling support option will have no effect.")
+  else()
+    # Noting to do for profiling on the host for now.
+    message(NOTICE "Profiling support enabled, but Re::Solve does not create tracer annotations for host code.")
+    message(NOTICE "This profiling support option will have no effect.")
```
cameronrutherford, line 154: Could support this with a different profiling tool like caliper/tau... At the same time, I don't see why we couldn't use ROCm/CUDA profilers to view host code performance as well, albeit that would require a hack + messy CMake

cameronrutherford (approved):

resolve/CMakeLists.txt
pelesh: You are right, we can, it is just that I didn't do it in this PR. Probably the message should say "tracer annotation not implemented for ...".

PR #135, opened 2024-02-07
Enhancements to Matrix Functions

### LINKED ISSUE #122: set*Data functions are dangerous
`set*Data` functions in linear algebra objects set external data to local data pointers within the object. However, they don't check if the pointer is null before assigning different address to it and therefore they are likely to cause memory leaks.

Consider checking whether pointer is null or not before assigning an address to it and implementing a remedy if it is not null.

pelesh:
Several enhancements to matrix operations including:
- Checks if data already exists when setting matrix and vector data. If so, report error.
- Export Re::Solve matrices to MatrixMarket file format.
- After deleting matrix data set data pointers to null.
- Add print functions to `Csr` and `Csc` matrix classes.
- Closes #122.
- Adds ILU0 factorization on CPU and appropriate unit test.
- Add hijacking constructor to CSR matrix class.


pelesh:
Code builds cleanly on x86_64 and AMD GPUs. All tests pass (run manually, CI pipelines are broken).

resolve/CMakeLists.txt
```diff
@@ -47,7 +48,7 @@ set(ReSolve_HEADER_INSTALL
     cusolver_defs.hpp
     LinSolver.hpp
     LinSolverIterativeFGMRES.hpp
-    RefactorizationSolver.hpp
+    LinSolverDirectCpuILU0.cpp
```
kswirydo, line 51: why do you have a cpp file on a header list???

resolve/LinSolverDirectCpuILU0.cpp
```diff
@@ -0,0 +1,349 @@
...
+  }
+
+  /**
+   * @brief Destructor
+   * 
+   * @todo Address how L and U factors are deleted (currently base class does that).
+   */
+  LinSolverDirectCpuILU0::~LinSolverDirectCpuILU0()
+  {
+    if (false) { //(owns_factors_) {
```
kswirydo, line 31: but this is never met, is it?

resolve/LinSolverDirectCpuILU0.cpp
```diff
@@ -0,0 +1,349 @@
...
+    // Use hijacking constructor to create L and U factors
+    L_ = new matrix::Csr(N, N, nnzL, false, true, &rowsL, &colsL, &valsL, memory::HOST, memory::HOST);
+    U_ = new matrix::Csr(N, N, nnzU, false, true, &rowsU, &colsU, &valsU, memory::HOST, memory::HOST);
+
+    owns_factors_ = true;
+
+    return error_sum;
+  }
+
+  int LinSolverDirectCpuILU0::reset(matrix::Sparse* A)
```
kswirydo, line 163: after you update values, can't you just call the setup again? thats a lot of repeated code (?)

resolve/matrix/Csr.cpp
```diff
@@ -37,6 +38,119 @@ namespace ReSolve
...
+    if ((memspaceSrc == memory::HOST)   && (memspaceDst == memory::HOST))  { control = 0;}
+    if ((memspaceSrc == memory::HOST)   && (memspaceDst == memory::DEVICE)){ control = 1;}
+    if ((memspaceSrc == memory::DEVICE) && (memspaceDst == memory::HOST))  { control = 2;}
+    if ((memspaceSrc == memory::DEVICE) && (memspaceDst == memory::DEVICE)){ control = 3;}
+
+    switch (control)
+    {
+      case 0: // cpu->cpu
+        // Set host data
+        h_row_data_ = *rows;
```
kswirydo, line 78: what is the purpose of double pointers here?

resolve/vector/Vector.cpp
```diff
@@ -91,21 +94,34 @@ namespace ReSolve { namespace vector {
...
    * @warning This function DOES NOT ALLOCATE any data, it only assigns the pointer.  
    */
-  void Vector::setData(real_type* data, memory::MemorySpace memspace)
+  int Vector::setData(real_type* data, memory::MemorySpace memspace)
   {
     using namespace ReSolve::memory;
     switch (memspace) {
       case HOST:
+        if (owns_cpu_data_ && h_data_) {
+          out::error() << "Trying to set vector host values, but the values already set!\n";
```
kswirydo, line 103: if you error out here, is the next line even printed?

kswirydo (commented):
Looks good, other than double pointers!!!

kswirydo:
When I am trying to compile, I am getting linker error
`[ 78%] Linking CXX executable version.exe
In file included from /people/swir617/ReSolve/tests/unit/matrix/runMatrixIoTests.cpp:7:
/people/swir617/ReSolve/tests/unit/matrix/MatrixIoTests.hpp: In member function ‘ReSolve::tests::TestOutcome ReSolve::tests::MatrixIoTests::cooMatrixExport()’:
/people/swir617/ReSolve/tests/unit/matrix/MatrixIoTests.hpp:89:37: error: ‘max_element’ is not a member of ‘std’
   89 |     const index_type N = 1 + *(std::max_element(rows.begin(), rows.end()));`

kswirydo:
LArge example works correctly.

resolve/LinSolverDirectCpuILU0.cpp
pelesh: This is temporary until deleting L and U factors is removed from the base class destructor. I will create a separate issue for that.

resolve/matrix/Csr.cpp
pelesh: The purpose of pointers to pointers is to set original pointers to data arrays to null. That way, the matrix object owns the data and there are no pointers outside the matrix that provide access to data arrays anymore.

resolve/vector/Vector.cpp
pelesh: Yes, it is. The `error()` here is just a buffer where error messages are streamed. It does not throw exception or interrupt code execution.

resolve/LinSolverDirectCpuILU0.cpp
pelesh: The repeated code is incomplete factorization. It is now pulled into `factorize` function. See [d48a737](https://github.com/ORNL/ReSolve/pull/135/commits/d48a737beeee12b5e6668cd8da0a913bc9e7567c)

resolve/LinSolverDirectCpuILU0.cpp
pelesh: To be tracked in #139.

maksud (approved):

PR #136, opened 2024-02-15
CPU-only iterative solvers

pelesh:
This is partial cleanup of @kswirydo's CPU-only branch, which enables selecting CPU-only iterative solvers in `SystemSolver` class. Before this PR, all iterative solvers required GPU support.

This is partial contribution to a feature branch; still far away from being ready to merge into `develop`.


kswirydo (commented):
looks good to me

PR #137, opened 2024-02-15
add cpu only rand solver to develop branch. 

kswirydo:
In this branch:
- cpu only randGMRES
- cpu only sketching methods

Rebasing is probably is required...

PR #138, opened 2024-02-16
Fix compiler warnings

pelesh:
Fixed most of compiler workings in #137.

### DIFF
--- a/resolve/LinSolverDirectSerialILU0.cpp
+++ b/resolve/LinSolverDirectSerialILU0.cpp
@@ -14,8 +14,8 @@ namespace ReSolve
 
   LinSolverDirectSerialILU0::~LinSolverDirectSerialILU0()
   {
-    delete h_aux1_;
-    delete h_ILU_vals_;
+    delete [] h_aux1_;
+    delete [] h_ILU_vals_;
   }
 
   int LinSolverDirectSerialILU0::setup(matrix::Sparse* A,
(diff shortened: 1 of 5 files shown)


resolve/LinSolverDirectSerialILU0.cpp
```diff
@@ -106,6 +106,7 @@ namespace ReSolve
         }
       }  
     }
+    // TODO: What is the purpose of nnzL and nnzU if they are not used after this?
```
kswirydo, line 109: ups thats a bug - now that I have nnzL and nnzU,  they should be used while allocating L and U!!! will fix after this is approved.

tests/unit/vector/GramSchmidtTests.hpp
```diff
@@ -165,7 +165,7 @@ namespace ReSolve {
               a->update(x->getVectorData(i, ms), ms, memory::HOST);
               b->update(x->getVectorData(j, ms), ms, memory::HOST);
               ip = handler->dot(a, b, memory::HOST);
-              if ( (i != j) && (abs(ip) > 1e-14)) {
+              if ( (i != j) && !isEqual(ip, 0.0)) {
```
kswirydo, line 168: if you use isEqual here, GramSchmidt test fails on deception (on CPU).

kswirydo (approved):

tests/unit/vector/GramSchmidtTests.hpp
kswirydo: although, there was one other bug that might had been causing this so I take this back (the bug had been fixed)

PR #141, opened 2024-02-21
working cgs1 orthogonalization in Gram-Schmidt

kswirydo:
Added CGS1 and a test for it.

### DIFF
--- a/tests/unit/vector/GramSchmidtTests.hpp
+++ b/tests/unit/vector/GramSchmidtTests.hpp
@@ -208,7 +208,7 @@ namespace ReSolve {
               if ( (i != j) && !isEqual(ip, 0.0, tol)) {
                 status = false;
                 std::cout << "Vectors " << i << " and " << j << " are not orthogonal!"
-                  << " Inner product computed: " << ip << ", expected: " << 0.0 << "\n";
+                          << " Inner product computed: " << ip << ", expected: " << 0.0 << "\n";
                 break; 
               }
               if ( (i == j) && !isEqual(sqrt(ip), 1.0, tol)) {           
--- a/tests/unit/vector/GramSchmidtTests.hpp
+++ b/tests/unit/vector/GramSchmidtTests.hpp
@@ -108,9 +108,11 @@ namespace ReSolve {
 
           GS->orthogonalize(N, V, H, 0); 
           GS->orthogonalize(N, V, H, 1); 
-
-          status *= verifyAnswer(V, 3, handler, memspace_);
-          
+          if (var ==  GramSchmidt::cgs1) {
+            status *= verifyAnswer(V, 3, handler, memspace_, 1e-13);
+          } else {
+            status *= verifyAnswer(V, 3, handler, memspace_);
+          }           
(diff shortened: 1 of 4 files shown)


tests/unit/vector/GramSchmidtTests.hpp
```diff
@@ -108,9 +108,11 @@ namespace ReSolve {
...
           GS->orthogonalize(N, V, H, 0); 
           GS->orthogonalize(N, V, H, 1); 
-
-          status *= verifyAnswer(V, 3, handler, memspace_);
-          
+          if (var ==  GramSchmidt::cgs1) {
+            status *= verifyAnswer(V, 3, handler, memspace_, 1e-13);
+          } else {
+            status *= verifyAnswer(V, 3, handler, memspace_);
+          }
```
pelesh, line 115: The magic number 1e-13 seems to be arbitrary. If we keep cgs1 for completeness only, knowing it is not stable, perhaps we should set this test to return "expected failure" status and use the same pass/fail criterions as for the other tests?

tests/unit/vector/GramSchmidtTests.hpp
```diff
@@ -168,7 +170,7 @@ namespace ReSolve {
               if ( (i != j) && !isEqual(ip, 0.0)) {
                 status = false;
                 std::cout << "Vectors " << i << " and " << j << " are not orthogonal!"
-                          << " Inner product computed: " << ip << ", expected: " << 0.0 << "\n";
+                  << " Inner product computed: " << ip << ", expected: " << 0.0 << "\n";
```
pelesh, line 173: Alignment of `<<`?

tests/unit/vector/GramSchmidtTests.hpp
```diff
@@ -183,6 +185,45 @@ namespace ReSolve {
...
+
+          for (index_type i = 0; i < K; ++i) {
+            for (index_type j = 0; j < K; ++j) {
+              a->update(x->getVectorData(i, ms), ms, memory::HOST);
+              b->update(x->getVectorData(j, ms), ms, memory::HOST);
+              ip = handler->dot(a, b, memory::HOST);
+              if ( (i != j) && !isEqual(ip, 0.0, tol)) {
+                status = false;
+                std::cout << "Vectors " << i << " and " << j << " are not orthogonal!"
+                  << " Inner product computed: " << ip << ", expected: " << 0.0 << "\n";
```
pelesh, line 211: Alignment of streaming operators:
```suggestion
                std::cout << "Vectors " << i << " and " << j << " are not orthogonal!"
                          << " Inner product computed: " << ip << ", expected: " << 0.0 << "\n";
```

pelesh (approved):
I would use uniform test criteria for all orthogonalization methods. If cgs1 does not pass, I would set that test to "expect failure".

tests/unit/vector/GramSchmidtTests.hpp
kswirydo: thats fine I just dont know how to do this.

PR #142, opened 2024-02-21
Let classical Gram-Schmidt fail expectedly

pelesh:
Classical Gram-Schmidt orthogonalization is numerical unstable. Rather than relaxing unit test passing criteria for this method, let us flag it as an expected failure. This will also signal users _not_ to use this method, which is provided in ReSolve for completeness purposes.

This is also a very good demo of `status.expectFailure()` feature in our unit testing framework 😄 .

### DIFF
--- a/tests/unit/TestBase.hpp
+++ b/tests/unit/TestBase.hpp
@@ -210,6 +210,8 @@ static const real_type half = 0.5;
 static const real_type one = 1.0;
 static const real_type two = 2.0;
 static const real_type three = 3.0;
+
+/// @brief eps = 2.2e-15 for double type
 static const real_type eps = 10*std::numeric_limits<real_type>::epsilon();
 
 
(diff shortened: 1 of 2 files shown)


pelesh:
CC @cameronrutherford @maksud

cameronrutherford (approved):
LGTM. Maybe if we had CDash configured we could get a better sense of how often this passes/fails

kswirydo (approved):
On GPU (nvidia), the CGS1 tests succeeds, so we see output like this
```
Test Summary
	Successful tests:     10
	Failed test:          0
	Skipped tests:        0
	Expected failures:    1
	Unexpected successes: 1
```

kswirydo:
> LGTM. Maybe if we had CDash configured we could get a better sense of how often this passes/fails

It fails on CPU but passes on nvidia GPU. Let me try hip too.

PR #143, opened 2024-02-22
Base class direct solver should not delete L and U factors

### LINKED ISSUE #139: Who should delete L and U factors in direct solver classes
Destructor of the base class `LinSolverDirect` deletes matrix factors L and U, which its members. However, L and U factors are created by classes derived from `LinSolverDirect`, hence it might be better if those classes have responsibility of deleting the factors.

pelesh:
Base class `LinSolverDirect` deletes member factors `L_` and `U_`, as well as permutation vectors `P_` and `Q_`, but it does not allocate them. It is better if the (derived) class that allocates these objects also assumes responsibility of deleting them.

Closes #139

### DIFF
--- a/resolve/LinSolverDirectCpuILU0.cpp
+++ b/resolve/LinSolverDirectCpuILU0.cpp
@@ -19,8 +19,6 @@ namespace ReSolve
   LinSolverDirectCpuILU0::LinSolverDirectCpuILU0(LinAlgWorkspaceCpu* /* workspace */)
     // : workspace_(workspace)
   {
-    std::cout << "LinSolverDirectCpuILU0 constructor\n";
-    std::cout << "L_ = " << L_ << ", P_ = " << P_ << "\n";
   }
 
   /**
@@ -30,9 +28,6 @@ namespace ReSolve
    */
   LinSolverDirectCpuILU0::~LinSolverDirectCpuILU0()
   {
-    std::cout << "LinSolverDirectCpuILU0 destructor\n";
-    std::cout << "L_ = " << L_ << ", P_ = " << P_ << "\n";
-    std::cout << "Deleting factors ...\n";
     if (owns_factors_) {
       delete L_;
       delete U_;
--- a/resolve/LinSolverDirectCpuILU0.cpp
+++ b/resolve/LinSolverDirectCpuILU0.cpp
@@ -19,6 +19,8 @@ namespace ReSolve
   LinSolverDirectCpuILU0::LinSolverDirectCpuILU0(LinAlgWorkspaceCpu* /* workspace */)
     // : workspace_(workspace)
   {
+    std::cout << "LinSolverDirectCpuILU0 constructor\n";
+    std::cout << "L_ = " << L_ << ", P_ = " << P_ << "\n";
   }
 
   /**
@@ -28,9 +30,14 @@ namespace ReSolve
    */
   LinSolverDirectCpuILU0::~LinSolverDirectCpuILU0()
   {
-    if (false) { //(owns_factors_) {
+    std::cout << "LinSolverDirectCpuILU0 destructor\n";
+    std::cout << "L_ = " << L_ << ", P_ = " << P_ << "\n";
+    std::cout << "Deleting factors ...\n";
+    if (owns_factors_) {
       delete L_;
       delete U_;
+      L_ = nullptr;
+      U_ = nullptr;
     }
     delete [] diagU_;
     delete [] idxmap_;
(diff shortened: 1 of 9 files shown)


resolve/LinSolverDirectCpuILU0.cpp
```diff
@@ -28,9 +28,11 @@ namespace ReSolve
    */
   LinSolverDirectCpuILU0::~LinSolverDirectCpuILU0()
   {
-    if (false) { //(owns_factors_) {
+    if (owns_factors_) {
```
kswirydo, line 31: where is the owns factors thing set? Is there a get/set method for it?

resolve/LinSolverDirectCpuILU0.cpp
```diff
@@ -28,9 +28,11 @@ namespace ReSolve
...
       delete L_;
       delete U_;
+      L_ = nullptr;
+      U_ = nullptr;
     }
     delete [] diagU_;
```
kswirydo, line 37: would this be not deleted, too, when the factors are owned by this class?

resolve/LinSolverDirectKLU.cpp
```diff
@@ -34,6 +34,16 @@ namespace ReSolve
 
   LinSolverDirectKLU::~LinSolverDirectKLU()
   {
+    if (factors_extracted_) {
```
kswirydo, line 37: where is this set?

resolve/LinSolverDirectSerialILU0.cpp
```diff
@@ -14,6 +14,12 @@ namespace ReSolve
 
   LinSolverDirectSerialILU0::~LinSolverDirectSerialILU0()
   {
+    if (owns_factors_) {
```
kswirydo, line 17: where is this set?

resolve/LinSolverDirectSerialILU0.cpp
```diff
@@ -109,8 +115,9 @@ namespace ReSolve
     // TODO: What is the purpose of nnzL and nnzU if they are not used after this?
     // allocate L and U
 
-    L_ = new matrix::Csr(n, n, nnz, false, true);  
-    U_ = new matrix::Csr(n, n, nnz, false, true);  
+    L_ = new matrix::Csr(n, n, nnz, false, true);
```
kswirydo, line 118: I think in one branch I pushed a version that uses nnzL here :D and nnzU in the next line. This should be updated.

kswirydo, line 118:
```suggestion
    L_ = new matrix::Csr(n, n, nnzL, false, true);
```

resolve/LinSolverDirectSerialILU0.cpp
```diff
@@ -109,8 +115,9 @@ namespace ReSolve
...
+    U_ = new matrix::Csr(n, n, nnz, false, true);
```
kswirydo, line 119:
```suggestion
    U_ = new matrix::Csr(n, n, nnzU, false, true);
```

kswirydo (commented):
The only comment i have is that we need to make sure factor ownership can be extracted and set (we have get/set functions). You can even add ows_factors to base class and add gets and sets there, it would make life easier (even though this class does not really use it). Not a clean solution but will work.

resolve/LinSolverDirectCpuILU0.cpp
pelesh: Boolean `owns_factors_` is set in `analyze()` function, which creates and allocates L and U factors. There are no accessor functions for this parameter as I don't anticipate anyone would need to change its value from outside the class.

resolve/LinSolverDirectSerialILU0.cpp
pelesh: On line 120, just after L and U are created.

resolve/LinSolverDirectCpuILU0.cpp
pelesh: Yes, it's redundant :)

resolve/LinSolverDirectSerialILU0.cpp
pelesh: Fixed in [cb42f10](https://github.com/ORNL/ReSolve/pull/143/commits/cb42f10faa1c4ed0112a497a56c2809f80e0a3a3)

resolve/LinSolverDirectSerialILU0.cpp
pelesh: Fixed in [cb42f10](https://github.com/ORNL/ReSolve/pull/143/commits/cb42f10faa1c4ed0112a497a56c2809f80e0a3a3)

resolve/LinSolverDirectSerialILU0.cpp
kswirydo: OK

kswirydo (approved):

PR #148, opened 2024-03-23
Ensure commutativity of GMRES solver configuration settings

pelesh:
The objective is to set GMRES configuration options in any order and get the same result. Currently, setting matrix must happen after all other configuration options are set. This PR includes:
- Function that changes Gram-Schmidt variant while preserving other solver settings.
- Function that changes sketching method while preserving other randomized GMRES settings.
- Method to change GMRES restart, while preserving all other settings.
- Function to switch between flexible and regular GMRES without affecting other solver settings.

All of these settings can take place in any desired order, before or after solver's `setup` method is called.


pelesh:
@kswirydo, would you think that Gram-Scmidt setup call should be made from inside of a GMRES class rather than outside of it? I would think a GMRES class should set up the Gram-Schmidt to its liking. It is hard to think of a scenario where the GS method would allocate its data to sizes other than what the calling GMRES requires.

maksud (approved):
All test cases were successful.

PR #151, opened 2024-06-03
initial implementation of #150

### LINKED ISSUE #54: Add `pre-commit` configuration + GitHub Actions
Things that pre-commit can help with:

- C++ lint
- CMake lint
- YAML validation
- Add JSON validation https://gist.github.com/jeekl/5083519

### LINKED ISSUE #150: Clarify style guidelines in the online documentation and set up `clang-format`
related to #54 / #82, albeit it deserves a dedicated pull request to unbundle it from the addition of `pre-commit`

# rationale

currently, [`CONTRIBUTING.md`](https://github.com/ORNL/ReSolve/blob/develop/CONTRIBUTING.md) is insufficiently precise in some ways and the repository's existing `.clang-format` file is not used and does not match the conventions within `CONTRIBUTING.md` exactly. clarifying the style guidelines and setting up `.clang-format` would help maintain a consistent style across the codebase and potentially make it more welcoming to newcomers

# considerations

- how is existing code formatted in relation to what is documented in `CONTRIBUTING.md`?
- how should the codified style be imposed upon existing code?
  - should we incrementally reformat code when it is touched, or should we reformat it all in one go?
- should we add a [precommit hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks)?
  - this would necessitate reformatting it all in one go, unless some work was done to detect modified files within the hook
  - or should we use [`pre-commit`](https://pre-commit.com/) instead, as described in #54?
 - what additional linting tools should we use and how/when should they be applied?

# initial suggestions

- drop #54 & #82 for now
- open a new pull request deleting the existing `.clang-format` file in `.github/workflows`[^1] since it isn't used and remove the dead code within the root `CMakeLists.txt`[^2] that references it
- create a new `.clang-format` at the repository's root which takes into account existing style conventions + the stuff within `CONTRIBUTING.md`, with a priority toward the latter, updating `CONTRIBUTING.md` accordingly
- do not apply the styling to existing code until a solution is agreed upon

[^1]: https://github.com/ORNL/ReSolve/blob/develop/.github/workflows/.clang-format
[^2]: https://github.com/ORNL/ReSolve/blob/d913854aed438af040f6ecd5bee052da3a861baa/CMakeLists.txt#L56-L58

superwhiskers:
this pull request implements the initial suggestions proposed within #150

i'm not going to merge this until some discussion about the formatting options in `.clang-format` has happened, and appropriate modifications have been made to `CONTRIBUTING.md`

Closes #54 
Closes #150

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -4,6 +4,7 @@ AlignConsecutiveMacros: true
 BreakBeforeBraces: Linux
 IndentWidth: 2
 ConstructorInitializerIndentWidth: 2
+ColumnLimit: 0
 AlignAfterOpenBracket: Align
 Standard: c++11
 IndentAccessModifiers: true
@@ -33,7 +34,7 @@ AllowShortFunctionsOnASingleLine: None
 AllowAllArgumentsOnNextLine: false
 IncludeBlocks: Regroup
 IncludeCategories:
-    - Regex: '^<resolve/'
+    - Regex: '^<(resolve|tests)/'
       Priority: 1
     - Regex: '^<[[:alnum:].]+>'
       Priority: -1
--- a/.clang-format
+++ b/.clang-format
@@ -20,13 +20,6 @@ PointerAlignment: Left
 InsertBraces: true
 NamespaceIndentation: All
 LineEnding: LF
-IntegerLiteralSeparator:
-    Binary: 4
-    BinaryMinDigits: 8
-    Decimal: 3
-    DecimalMinDigits: 4
-    Hex: 2
-    HexMinDigits: 6
 AlwaysBreakTemplateDeclarations: true
 BinPackArguments: false
 BinPackParameters: false
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,41 @@
+Language: Cpp
+AccessModifierOffset: 0
+AlignConsecutiveMacros: true
+BreakBeforeBraces: Linux
+IndentWidth: 2
+ConstructorInitializerIndentWidth: 2
+AlignAfterOpenBracket: Align
+Standard: c++11
+IndentAccessModifiers: true
+IndentCaseBlocks: true
+IndentCaseLabels: true
+IndentGotoLabels: true
+IndentExternBlock: Indent
+UseTab: Never
+ReflowComments: true
+QualifierAlignment: Left
+ReferenceAlignment: Pointer
+PointerAlignment: Left
+InsertBraces: true
+NamespaceIndentation: All
+LineEnding: LF
+IntegerLiteralSeparator:
+    Binary: 4
+    BinaryMinDigits: 8
+    Decimal: 3
+    DecimalMinDigits: 4
+    Hex: 2
+    HexMinDigits: 6
+BreakTemplateDeclarations: Yes
+BinPackArguments: false
+BinPackParameters: false
+AllowShortFunctionsOnASingleLine: None
+AllowAllArgumentsOnNextLine: false
+IncludeBlocks: Regroup
+IncludeCategories:
+    - Regex: '^<resolve/'
+      Priority: 1
+    - Regex: '^<[[:alnum:].]+>'
+      Priority: -1
+    - Regex: '.*'
+      Priority: 0
--- a/.clang-format
+++ b/.clang-format
@@ -1,35 +1,105 @@
-Language: Cpp
-AccessModifierOffset: 0
-AlignConsecutiveMacros: true
-BreakBeforeBraces: Linux
-IndentWidth: 2
+Language:        Cpp
+BasedOnStyle:  Microsoft
+
+# Indentation
+IndentWidth: 2                 # 2 spaces per indent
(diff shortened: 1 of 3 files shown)


.clang-format
```diff
@@ -0,0 +1,41 @@
...
+IndentExternBlock: Indent
+UseTab: Never
+ReflowComments: true
+QualifierAlignment: Left
+ReferenceAlignment: Pointer
+PointerAlignment: Left
+InsertBraces: true
+NamespaceIndentation: All
+LineEnding: LF
+IntegerLiteralSeparator:
```
cameronrutherford, line 22: https://stackoverflow.com/questions/27767781/why-was-the-space-character-not-chosen-for-c14-digit-separators - I think this is just a c++14 feature, which won't work for Re::Solve?

.clang-format
```diff
@@ -0,0 +1,41 @@
...
+    Binary: 4
+    BinaryMinDigits: 8
+    Decimal: 3
+    DecimalMinDigits: 4
+    Hex: 2
+    HexMinDigits: 6
+AlwaysBreakTemplateDeclarations: true
```
cameronrutherford, line 29: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alwaysbreaktemplatedeclarations - This should just be `BreakTemplateDeclarations`

```suggestion
BreakTemplateDeclarations: Yes
```

https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breaktemplatedeclarations - I think we should add something about Templates to `CONTRIBUTING.md`, although I don't think we rely on templates in many places in Re::Solve

cameronrutherford (approved):
A few minor comments, but I approve.

I am not a regex expert so I did not have a chance to verify that those are correct.

.clang-format
superwhiskers: that's correct. i added that prior to getting the tentative suggestion to not increase the c++ version from @pelesh, but left it in there. i'll drop it if we come to a consensus to not increase the version on this pull request

.clang-format
superwhiskers: [i deliberately reverted it to the deprecated form](https://github.com/ORNL/ReSolve/pull/151/commits/3f61b211a1b376171d740a026e359287a16cfacc) because my local clang-format version is older than 19

.clang-format
```diff
@@ -0,0 +1,41 @@
...
+AlignConsecutiveMacros: true
+BreakBeforeBraces: Linux
+IndentWidth: 2
+ConstructorInitializerIndentWidth: 2
+AlignAfterOpenBracket: Align
+Standard: c++11
+IndentAccessModifiers: true
+IndentCaseBlocks: true
+IndentCaseLabels: true
+IndentGotoLabels: true
```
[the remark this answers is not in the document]
superwhiskers: i think most people are going to avoid using it anyway. there are a handful of legitimate uses for it anyway, such as exiting nested loops, and in c, sharing cleanup code prior to exiting a function. it's probably not worth explicitly forbidding it

.clang-format
cameronrutherford: I doubt we will change the C++ standard in this PR, but we should in the future

.clang-format
cameronrutherford: Sounds like we should specify what `clang-format` version we support / expect

superwhiskers: a solid minimum would be 16. some of the other features i'm using in the `.clang-format` here were added in 16 and that's the version i'm using locally

.clang-format
```diff
@@ -0,0 +1,41 @@
+Language: Cpp
+AccessModifierOffset: 0
+AlignConsecutiveMacros: true
+BreakBeforeBraces: Linux
+IndentWidth: 2
+ConstructorInitializerIndentWidth: 2
+AlignAfterOpenBracket: Align
+Standard: c++11
```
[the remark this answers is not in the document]
pelesh: My thinking is we should bump to C++17 standard when there is a compelling reason to do so. I suggest we move such discussion to a separate issue.

pelesh:
Before it is ready for merge, we need to test it on a few representative code snippets to make sure there are no unintended artifacts.

.clang-format
```diff
@@ -0,0 +1,41 @@
+Language: Cpp
+AccessModifierOffset: 0
+AlignConsecutiveMacros: true
+BreakBeforeBraces: Linux
+IndentWidth: 2
+ConstructorInitializerIndentWidth: 2
```
pelesh, line 6: It seems that the line with specification is missing here compared to the earlier specs. I suggest setting it to "0" to disable automatic line wrapping because no matter how you specify those, automatic wrapping will eventually give you undesired effect. Too long lines of code can be addressed in PR reviews easily.
```suggestion
ColumnLimit: 0
ConstructorInitializerIndentWidth: 2
```

superwhiskers:
dropping in favor of #261

PR #163, opened 2024-06-13
Parameter management utilities

### LINKED ISSUE #96: Comprehensive way to manage ReSolve configuration parameters
ReSolve includes several different linear solvers and supports three hardware backends. There is a large number of parameters in ReSolve to select specific solution strategy and to tune its performance. Many of the parameters are specific to a specific solver type or hardware backend or both. We need a hierarchical method to allow user to configure ReSolve easily without sifting through dozens of (often cryptically described) parameters. Most of these parameters need to be set at runtime.

This issue is to start discussion how to handle this most effectively.

The issue originally raised by @kswirydo in https://github.com/ORNL/ReSolve/pull/86#discussion_r1412715592_

pelesh:
Managing different solver parameters in Re::Solve becomes a nontrivial task and there is a need for consistent parameter management across all Re::Solve modules. In addition to setting up file I/O for initializing solver parameters, the scope here includes creating a common interface for setting and retrieving parameters for all solver classes and ability to change solver configuration parameters at runtime. The requirements are:
- [x] Create a generic parameter management interface at the base class, which each derived solver class can populate with its own specific parameters.
- [x] Parameters can be of different types (integer, floating point, string, etc.).
- [x] This is core functionality, no additional dependency should be required to implement it.
- [x] Changing parameters should leave solver in fully functional state with new parameter values.

This PR is a proposal how to handle this. Parameter management is implemented only for classes `LinSolverIterativeFGMRES` and `LinSolverIterativeRandGMRES`. After this is agreed upon and merged, parameter management will be added to other solvers.


resolve/LinSolver.hpp
```diff
@@ -34,94 +48,44 @@ namespace ReSolve
       virtual ~LinSolver();
 
       real_type evaluateResidual();
+
+      virtual int setCliParam(const std::string /* id */, const std::string /* value */)
+      {
+        return 1;
+      }
```
cameronrutherford, line 55: I have a personal preference for these functions being one line, but that has been discussed in #151

tests/functionality/testSysRandGMRES.cpp
```diff
@@ -129,17 +129,23 @@ int test(int argc, char *argv[])
   real_type norm_b = 0.0;
 
   // Set solver options
-  real_type tol = 1e-12;
+  //real_type tol = 1e-12;
```
cameronrutherford, line 132:
```suggestion
```

tests/functionality/testSysRandGMRES.cpp
```diff
@@ -129,17 +129,23 @@ int test(int argc, char *argv[])
...
   solver.getIterativeSolver().setMaxit(2500);
-  solver.getIterativeSolver().setTol(tol);
+  // solver.getIterativeSolver().setTol(tol);
```
cameronrutherford, line 134:
```suggestion
```

tests/functionality/testSysRandGMRES.cpp
```diff
@@ -129,17 +129,23 @@ int test(int argc, char *argv[])
...
+  solver.getIterativeSolver().getCliParam("tol", tol);
 
   matrix_handler.setValuesChanged(true, memspace);
 
   // Set system matrix and initialize iterative solver
   status = solver.setMatrix(A);
   error_sum += status;
 
-  solver.getIterativeSolver().setRestart(200);
+  // solver.getIterativeSolver().setRestart(200);
```
cameronrutherford, line 145:
```suggestion
```

resolve/LinSolverIterativeRandFGMRES.cpp
```diff
@@ -502,6 +505,109 @@ namespace ReSolve
...
+  int LinSolverIterativeRandFGMRES::getCliParam(const std::string id, std::string& /* value */)
+  {
+    switch (params_list_[id])
+    {
+      default:
+        out::error() << "Unknown string parameter " << id << "\n";
+        return 1;
+    }
+    return 0;
+  }
```
cameronrutherford, line 541: I am a little confused about the purpose of this function. Is that to say no parameters have a string as their value?

resolve/LinSolverIterativeFGMRES.cpp
```diff
@@ -143,18 +146,31 @@ namespace ReSolve
...
+      //   exit_cond =  ((std::abs(rnorm - ZERO) <= EPSILON));
+      // } else {
+      //   if (conv_cond_ == 1) {
+      //     exit_cond =  ((std::abs(rnorm - ZERO) <= EPSILON) || (rnorm < tol_));
+      //   } else {
+      //     if (conv_cond_ == 2) {
+      //       exit_cond =  ((std::abs(rnorm - ZERO) <= EPSILON) || (rnorm < (tol_*bnorm)));
+      //     }
+      //   }
+      // }
```
cameronrutherford, line 173: Should probably remove this if it's unnecessary as the switch statement is cleaner.

cameronrutherford (commented):
I don't really have substantial feedback...

tests/functionality/testSysRandGMRES.cpp
pelesh: Good catch!

resolve/LinSolverIterativeRandFGMRES.cpp
pelesh: You are right. The base class should implement parameter getters returning error messages. Derived classes should implement parameter getters only for parameters that are defined for that specific solver.

resolve/LinSolverIterativeFGMRES.cpp
pelesh: Thanks!

stonecoldhughes:
https://github.com/ORNL/ReSolve/issues/96#issuecomment-2362551756

stonecoldhughes:
https://en.cppreference.com/w/cpp/language/raii is an idea that applies to configuration - ideally, an object is completely configured and ready to use as soon as it is constructed. Are these configurable solver objects configured with the parsed configuration options in their constructors? Or is the functionality here used to set that up after the fact?

pelesh:
> https://en.cppreference.com/w/cpp/language/raii is an idea that applies to configuration - ideally, an object is completely configured and ready to use as soon as it is constructed. Are these configurable solver objects configured with the parsed configuration options in their constructors? Or is the functionality here used to set that up after the fact?

Since Re::Solve is used for solving series of similar linear systems it is conceivable that one might want to change its parameters at runtime. In that case, Re::Solve would need to ensure that the choice of solver configuration parameters is self-consistent and that correct workspace is allocated.

resolve/LinSolver.hpp
pelesh: To be tracked in #151.

resolve/LinSolverIterativeFGMRES.cpp
pelesh: Removed in [5356946](https://github.com/ORNL/ReSolve/pull/163/commits/535694632e9ebc197475d5649ec249c310b134c1).

stonecoldhughes (approved):
These changes are a big step in the right direction. According to https://en.wikipedia.org/wiki/Single-responsibility_principle, it is usually best to map each separate definable functionality to a separately testable object. The LinSolver base class appears to handle 3 things: parameter management, norm/residual calculation, and matrix equation solution. Is it possible for LinSolver to offload the first two into other classes to a greater degree? Would that increase the composability of the design in a desirable way?

pelesh:
I suggest we keep this as draft for now. We need to merge #198 and probably do some other fixes as we discover them through this preview PR. Thanks all for useful feedback so far.

tests/functionality/testSysRandGMRES.cpp
pelesh: Fixed in [8e1bd44](https://github.com/ORNL/ReSolve/pull/163/commits/8e1bd44561ae39bf6888da14d5c783681fe01bd2).

tests/functionality/testSysRandGMRES.cpp
pelesh: Fixed in [8e1bd44](https://github.com/ORNL/ReSolve/pull/163/commits/8e1bd44561ae39bf6888da14d5c783681fe01bd2).

resolve/LinSolver.hpp
```diff
@@ -34,94 +48,44 @@ namespace ReSolve
...
+        
+      virtual int getCliParam(const std::string /* id */, std::string& /* value */)
+      {
+        return 1;
+      }
+        
+      virtual int getCliParam(const std::string /* id */, index_type& /* value */)
```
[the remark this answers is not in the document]
pelesh: They can and they probably should. In this PR, we implement parameter management only for a couple of solver classes, just to see if this design is feasible. Once all solver classes implement these methods, we'll turn base class methods pure virtual.

tests/functionality/testSysRandGMRES.cpp
pelesh: Fixed in [8e1bd44](https://github.com/ORNL/ReSolve/pull/163/commits/8e1bd44561ae39bf6888da14d5c783681fe01bd2).

pelesh:
> https://en.cppreference.com/w/cpp/language/raii is an idea that applies to configuration - ideally, an object is completely configured and ready to use as soon as it is constructed. Are these configurable solver objects configured with the parsed configuration options in their constructors? Or is the functionality here used to set that up after the fact?

This is a very good point. In sparse numerical linear algebra, one needs to compute the size of the workspace _after_ all the objects are instantiated. Similarly, changing some of the configuration parameters requires recomputing and reallocating the workspace. Therefore, solver object typically is _not_ ready to use as soon as it is constructed.

tests/functionality/testSysRandGMRES.cpp
```diff
@@ -177,7 +198,7 @@ int test(int argc, char *argv[])
     std::cout << "Result is not a finite number!\n";
     error_sum++;
   }
-  if (final_norm/norm_b > (10.0 * tol)) {
+  if (final_norm/norm_b > (10.0 * tol_out)) {
```
shakedregev, line 201: Maybe this `10.0` should be an optional parameter? I can imagine a scenario where I specify a small tolerance, but don't necessarily want to throw an error if I wasn't able to achieve it (or 10X it)

shakedregev (approved):
I pulled and ran the new test, it works. The code is understandable, but we should be able to find a way to make these MRs more manageable, as much of the code was repeated.

tests/functionality/testSysRandGMRES.cpp
pelesh: See #203, for example. We do need to make these tests more modular and more configurable.

maksud (approved):
Looks good to me based on the discussions.

Issue #155, opened 2024-06-11
Add function to expand symmetric matrix stored as an upper/lower triangular to a full square matrix


pelesh:
In some instances an application generating symmetric matrices in a lower/upper triangle storage format may use Re::Solve's LU solvers. In that case, the matrix needs to be expanded to the full square matrix format. It would be helpful to have `expandSymmetricMatrix` function available to make such transformation when needed.

First to decide is where this function should be implemented:
- As a member of `matrix::Sparse` class?
- As a member of `matrix::MatrixHandler`?
- As a standalone function in `matrix` namespace.

Several implementation details that need to be decided:
- When to allocate device memory for the expanded matrix?
- What operations to carry out on the host and what on the device?
- How to reuse sparsity pattern of the expanded matrix over several linear solver calls?

CC @kswirydo @maksud

superwhiskers:
this will be closed by the merge of #164 unless the addition of `ReSolve::matrix::expand` is dropped.

superwhiskers:
actually, i suppose that doesn't totally resolve the issue since `ReSolve::matrix::expand` is currently only host-side. it still needs to be decided whether there's a need for gpu-related functionality here

pelesh:
This issue has been resolved in #184. Feel free to reopen it, if needed.

PR #164, opened 2024-06-14
add a LUSOL functionality test

superwhiskers:
this pull request adds a functionality test for LUSOL. there is one thing that needs to be worked out before this can actually be merged:

- how do we finish implementing symmetric matrix expansion functionality? this is necessary for the integration test to work since LUSOL cannot operate on symmetric matrices stored only as a half (#155)
  - currently, i've added a single method to the `matrix::Coo` class which expands the matrix. this is not permanent and it should be fully expanded upon prior to a merge with equivalent functionality for other matrix types


superwhiskers:
oh. this also adds comments referencing stuff about extracting L/U factors. i'll remove that so this can be focused only on testing

resolve/matrix/MatrixHandlerCpu.cpp
```diff
@@ -70,6 +71,42 @@ namespace ReSolve {
         sum *= (*alpha);
         result_data[i] = result_data[i]*(*beta) + sum;
       } 
+      vec_result->setDataUpdated(memory::HOST);
+      return 0;
+    } else if (matrixFormat == "coo") {
+      matrix::Coo* A = static_cast<matrix::Coo*>(Ageneric);
+      index_type* rows = A->getRowData(memory::HOST);
```
pelesh, line 78: Implement matvec for COO matrices as a static function in the functionality test for now.

resolve/matrix/Sparse.hpp
```diff
@@ -58,6 +58,8 @@ namespace ReSolve { namespace matrix {
 
       virtual int copyData(memory::MemorySpace memspaceOut) = 0;
 
+      int setDataOwnership(bool, memory::MemorySpace);
+      int setValueOwnership(bool, memory::MemorySpace);
```
pelesh, line 62: These are too dangerous methods and we should not implement them. I suggest implementing a [hijacking constructor](https://github.com/ORNL/ReSolve/blob/develop/resolve/matrix/Csr.cpp#L42) as in `Csr` class instead.

tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,167 @@
...
+  std::unique_ptr<ReSolve::VectorHandler> vector_handler(new ReSolve::VectorHandler(workspace.get()));
+
+  std::string matrix_one_path = data_path + "data/matrix_ACTIVSg200_AC_10.mtx";
+  std::string matrix_two_path = data_path + "data/matrix_ACTIVSg200_AC_11.mtx";
+
+  std::string rhs_one_path = data_path + "data/rhs_ACTIVSg200_AC_10.mtx.ones";
+  std::string rhs_two_path = data_path + "data/rhs_ACTIVSg200_AC_11.mtx.ones";
+
+  real_type scaled_residual_norm_one = compute_error(error_sum, matrix_handler, vector_handler, matrix_one_path, rhs_one_path);
+  real_type scaled_residual_norm_two = compute_error(error_sum, matrix_handler, vector_handler, matrix_two_path, rhs_two_path);
```
pelesh, line 153: If you want to do something like this, then the same solver instance has to be passed as the argument to both `compute_error` functions. Otherwise, there is little reason to run tests on two matrices. The `analysis` done at the first call needs to be reused in the second call. I suggest dropping `compute_error` function from this PR and handling it in a separate PR and across all functionality tests.

resolve/matrix/io.cpp
```diff
@@ -44,16 +44,26 @@ namespace ReSolve { namespace io {
...
+      if (a == b) {
+        n_diagonal++;
+      }
+
       i++;
     }
     A->setMatrixData(coo_rows, coo_cols, coo_vals, memory::HOST);
+    A->setDataOwnership(true, memory::HOST);
+    A->setValueOwnership(true, memory::HOST);
+    A->setNnzExpanded((2 * nnz) - n_diagonal);
```
pelesh, line 66: Use hijacking constructor here.

pelesh (changes requested):
A good first stab. Here are some suggestions how to proceed.

tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,167 @@
...
+  if (status != 0) {
+    error_sum += status;
+    return 1;
+  }
+
+  std::unique_ptr<vector_type> vec_test(new vector_type(A->getNumRows()));
+  std::unique_ptr<vector_type> vec_diff(new vector_type(A->getNumRows()));
+
+  std::unique_ptr<real_type[]> x_data(new real_type[A->getNumRows()]);
+  std::fill_n(x_data.get(), A->getNumRows(), 1.0);
```
pelesh, line 78: I think it is better if we use raw pointers for raw data as we want to move it back and forth from GPUs. 
```suggestion
  real_type* x_data = new real_type[A->getNumRows()];
  std::fill_n(x_data, A->getNumRows(), 1.0);
```

pelesh: We need to make it explicit in the code that raw data should be accessed through raw pointers. This is a CPU-only test, so in this case it does not matter but as the best practice, we should not use unique pointers for the raw matrix and vector data.

resolve/matrix/Utilities.cpp
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  int matrix::expand(matrix::Csr& A)
+  {
+    // NOTE: this is pretty abuse-resistant, but note that if you have (for
+    //       some reason) stored differing values on opposing sides of the
+    //       diagonal, the result is nominally undefined (but technically
+    //       predictable
+    //
+    //       ...please don't rely on it being predictable
```
pelesh, line 92: It is unclear to me what this means. In any case, please add notes comments and pre/post conditions in Doxygen style to the new functions.

pelesh (changes requested):
There are some minor issues that I addressed in my comments.

Newly added `expand()` function needs unit test(s). I suggest we create `MatrixConversionTest.hpp` and `runMatrixConversionTest.cpp` in `tests/unit` directory and implement needed unit tests there.

resolve/matrix/Utilities.cpp
```diff
@@ -0,0 +1,144 @@
...
+          n_diagonal++;
+        }
+      }
+
+      index_type nnz_expanded = (2 * nnz) - n_diagonal;
+      A.setNnzExpanded(nnz_expanded);
+
+      // NOTE: so because most of the code here uses new/delete and there's no
+      //       realloc(3) equivalent for that memory management scheme, we
+      //       have to manually new/memcpy/delete, unfortunately
```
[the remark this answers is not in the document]
superwhiskers: the note does not make any sense removed from the specific place of the source code it's located in

resolve/matrix/Sparse.hpp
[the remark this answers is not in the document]
superwhiskers: see #167

pelesh (changes requested):
Good progress but some more fixes are needed before this is ready to merge. I left a few comments to that end.

LUSOL test produces some type conversion warnings that need to be fixed.

resolve/matrix/Utilities.hpp
```diff
@@ -0,0 +1,26 @@
...
+    int expand(Coo&);
+
+    /**
+     * @brief Expands a symmetric CSR matrix stored only as a half to its full size
+     */
+    int expand(Csr&);
+
+    /**
+     * @brief Expands a symmetric CSC matrix stored only as a half to its full size
+     */
```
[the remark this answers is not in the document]
superwhiskers: no assumption is made. technically, you could pass in a matrix that isn't triangular and it would still work

resolve/matrix/Sparse.cpp
```diff
@@ -236,6 +236,40 @@ namespace ReSolve {
...
+      default:
+        return -1;
+    }
+
+    return 0;
+  }
+
+  /// @brief Set ownership of CPU/GPU values
+  int matrix::Sparse::setValueOwnership(bool owns, memory::MemorySpace memspace)
+  {
```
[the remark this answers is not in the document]
superwhiskers: i don't really see how that's any better. `setValueOwnership` is just as correct in a plural context

superwhiskers:
rebased on latest `develop`

cameronrutherford:
https://github.com/ORNL/ReSolve/blob/lusol-integration-tests/examples/CMakeLists.txt#L140 - might be good to have Lusol be used in the consumer test for testing `resolve~cuda~rocm~klu+lusol`

pelesh:
> https://github.com/ORNL/ReSolve/blob/lusol-integration-tests/examples/CMakeLists.txt#L140 - might be good to have Lusol be used in the consumer test for testing `resolve~cuda~rocm~klu+lusol`

I agree.  To get there, we need to get LUSOL to work within a refactorization solver strategy, which is outside the scope of this PR.

resolve/matrix/Utilities.hpp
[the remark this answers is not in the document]
pelesh: The function should not allow that. If the matrix is not upper- or lower-triangular, this function should return an error. If for some reason non-symmetric matrix was tagged as symmetric, that is a nasty bug, which we need to protect against.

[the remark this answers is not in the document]
superwhiskers: it's specified as part of the preconditions for each of the expand functions. i can add a debug check that goes away in release builds to catch any bugs that violate it, but i don't see why this should be a general check we do, as the presence of a precondition makes it on the callee to know beforehand

pelesh:
Closing in favor of #189.

PR #170, opened 2024-06-21
Add make test_install to github tests.

pelesh:
Running `ctest` triggers testing the build. Running `make test_install` starts installation tests. The latter will verify if all runtime paths are set correctly.


cameronrutherford (approved):
CI logs look good! Thanks for adding this

cameronrutherford:
I have a few ideas to speed up the spack pipelines:

1. only re-build base image if it doesn't exist
1. cache a spack.lock so we don't have to concretize

1 is definitely a no brainer, and should be done.

2 is a little more complicated. You would need to remember to remove that file when you update the spack submodule. Additionally we would have to maintain a lock file for each build in the matrix.

PR #172, opened 2024-06-24
Update spack_cpu_build.yaml to build ~klu+lusol

cameronrutherford:
Targeting #164 since there is more testing in that PR.

We could also merge into develop first, but merging into #164 just in cases tests fail...

cc @superwhiskers and @pelesh


pelesh (approved):
Yes, `~klu+lusol` configuration is needed.

cameronrutherford:
Happy to merge...

Issue #166, opened 2024-06-19
document invariants for `nnz_expanded_` or remove it entirely


superwhiskers:
according to @pelesh, there is no hard requirement that the value stored in the `nnz_expanded_` field on the `Sparse` class be correct. the field should either be removed entirely (as i'm not really sure what it would be used for if it can't be relied on) or have this invariant documented so that it does not lead to confusion or errors in the future

pelesh:
Thanks for reporting this issue, @superwhiskers.

The `nnz_expanded_` refers to the number of non zeros when you expand a symmetric matrix to a general sparse matrix format. There is no way of knowing `nnz_expanded_` upon instantiation of a sparse symmetric matrix in upper/lower triangular form, so it should be set to zero and stay zero for al long as flag `is_expanded_` is set to false. The `nnz_expanded_` needs to be calculated at the time when a function to expand the matrix into a general form is called. The expand function should have the responsibility to set the new number of nonzeros and flip the flag `is_expanded_` to true.

My preference would be to remove `nnz_expanded_` as a member variable altogether because it is used only as a temporary storage until `nnz_` is updated.

CC @kswirydo

PR #173, opened 2024-06-25
Remove duplicated code and consolidate coo2csr functionality in a single function.

pelesh:
Consolidate `coo2csr` functions in `MatrixHandler` and `Csr` classes into a single standalone function in `matrix` namespace. This utility function is now called by both, `MatrixHandler` and `Csr` classes.

This is first step before refactoring `coo2csr` function, as well as `matrix::Sparse` class to address #166.


superwhiskers:
without suggesting any additional modifications to the code, i ask that you move the `utilities.cpp` & `utilities.hpp` files to `Utilities.cpp` and `Utilities.hpp`, as in #164.

kswirydo (approved):

PR #174, opened 2024-07-05
Rework of coo2csr conversion

pelesh:
The new implementation of `coo2csr` has complexity $`O(NNZ\log(NNZ))`$. It handles unordered COO with duplicates. More testing is needed.

### DIFF
--- a/resolve/matrix/Utilities.cpp
+++ b/resolve/matrix/Utilities.cpp
@@ -1,4 +1,5 @@
 #include <algorithm>
+#include <list>
 #include <resolve/Common.hpp>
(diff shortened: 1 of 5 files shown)


superwhiskers:
my only comment to make here would be that there's #175 which i can open a pull request to the branch of to add in the fixes for the conversion code i developed while fixing and making #178

pelesh:
> my only comment to make here would be that there's #175 which i can open a pull request to the branch of to add in the fixes for the conversion code i developed while fixing and making #178

Tests fail in #175 so it is not ready to merge solution. I need to address #176 before we can implement a sustainable solution.

PR #175, opened 2024-07-05
alternate coo2csr implementation

superwhiskers:
this is a new coo2csr implementation, similar to the work done in #174, but written entirely from scratch, presumably with better performance (space/runtime---i've yet to benchmark it).

things that need to be done before this can be merged:
- [ ] merge #164 so we can resolve a `TODO` in the source code relating to a memory leak
- [ ] work out why the `klu_klu_test` fails (this is also a problem with #174)
- [ ] figure out how we can avoid filling the column data array with `-1`s
- [ ] benchmark it?


pelesh (changes requested):
This is an incomplete review catching a few places that may be bugs and suggesting how to make the code more readable.

superwhiskers:
all of the comments above should be addressed now

PR #178, opened 2024-07-19
lusol factor extraction

### LINKED ISSUE #159: Add functions to extract L and U factors from LUSOL
Implement following methods in `LinSolverDirectLUSOL` class:
- `getLFactor` and `getUFactor`. Each method should return `matrix::csc` object.
- `getPOrdering` and `getQOrdering`. Methods should return arrays of lengths `n_` and `m_`, respectively.

Implement unit tests for each method.

CC @kswirydo

superwhiskers:
Resolves #159.

what remains to be done:
- [x] rebase on `develop` from `lusol-integration-tests`
- [x] use already existing `L_` and `U_` member variables instead of allocating new ones
- [x] constrain size by nsing
- [x] add tests for factor extraction


superwhiskers:
wasn't able to solve it with a rebase. i ended up making a new branch locally and cherry-picking the commits from the old one

superwhiskers:
this should be ready to merge now

resolve/LinSolverDirectLUSOL.cpp
```diff
@@ -208,30 +213,203 @@ namespace ReSolve
...
+      return nullptr;
+    }
+
+    if (U_ != nullptr) {
+      // likewise
+      return U_;
+    }
+
+    U_ = static_cast<matrix::Sparse*>(new matrix::Csr(n_, m_, luparm_[23] - luparm_[10], false, true));
+    U_->allocateMatrixData(memory::HOST);
```
pelesh, line 323: This needs to be better documented. It is not obvious what are luparams 10 and 23.

resolve/LinSolverDirectLUSOL.cpp
```diff
@@ -208,30 +213,203 @@ namespace ReSolve
...
+
+    // fill the destination arrays
+
+    for (index_type row = 0; row < n_; row++) {
+      index_type offset = locr_[p_[row] - 1] - 1;
+
+      for (index_type destination_offset = rows[row];
+           destination_offset < rows[row + 1];
+           destination_offset++) {
+        index_type column = qt[indr_[offset] - 1];
```
pelesh, line 357:
```suggestion
      for (index_type destination_offset = rows[row]; destination_offset < rows[row + 1]; destination_offset++) {
        index_type column = qt[indr_[offset] - 1];
```

resolve/LinSolverDirectLUSOL.cpp
```diff
@@ -208,30 +213,203 @@ namespace ReSolve
...
+        // as said above, i'm pretty certain lusol doesn't write duplicates
+        if (columns[insertion_offset] == column) {
+          out::error() << "duplicate element found during LUSOL U factor extraction\n";
+          return nullptr;
+        }
+
+        for (index_type swap_offset = destination_offset;
+             swap_offset > insertion_offset;
+             swap_offset--) {
+          std::swap(columns[swap_offset], columns[swap_offset - 1]);
```
pelesh, line 375:
```suggestion
        for (index_type swap_offset = destination_offset; swap_offset > insertion_offset; swap_offset--) {
          std::swap(columns[swap_offset], columns[swap_offset - 1]);
```

pelesh (changes requested):
The code needs to be better commented. There are some nontrivial indexing patterns and there need to be comments explaining what is being done in main loops in L and U factor getters.

pelesh:
This is very good work. Apart from commenting code better, this should be ready to merge.

resolve/LinSolverDirectLUSOL.cpp
```diff
@@ -208,30 +213,203 @@ namespace ReSolve
...
+
+    index_type diagonal_bound = std::min({m_, n_});
+
+    L_ = static_cast<matrix::Sparse*>(new matrix::Csc(n_, m_, luparm_[22] + diagonal_bound, false, true));
+    L_->allocateMatrixData(memory::HOST);
+
+    index_type* columns = L_->getColData(memory::HOST);
+    index_type* rows = L_->getRowData(memory::HOST);
+    std::fill_n(rows, luparm_[22], -1);
+    real_type* values = L_->getValues(memory::HOST);
```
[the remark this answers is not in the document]
superwhiskers: we have to initialize it to something nonzero for the same reason the new coo2csr implementation (#175) does that. until i've worked out why exactly this works and not zero, this is the only reasonable choice, as any positive value risks conflict with actual column indices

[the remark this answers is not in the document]
pelesh: I think it would be good to investigate if we could implement this with the values of `rows` array initialized to zero, because the logic as implemented now would not work if `index_type` is unsigned integer.

pelesh (approved):

resolve/LinSolverDirectLUSOL.cpp
```diff
@@ -206,32 +228,218 @@ namespace ReSolve
...
+      columns[corresponding_column + 1] = column_nnz;
+      offset = column_nnz_end;
+    }
+
+    for (index_type column = 0; column < m_; column++) {
+      columns[column + 1] += columns[column];
+    }
+
+    // handle rectangular l factors correctly
+    for (index_type column = 0; column < diagonal_bound; column++) {
```
kswirydo, line 288: is this supposed to sort columns through bubble sort type algorithm?

resolve/LinSolverDirectLUSOL.cpp
```diff
@@ -333,20 +536,24 @@ namespace ReSolve
...
-    lenc_  = nullptr; 
-    lenr_  = nullptr;
-    locc_  = nullptr;
-    locr_  = nullptr; 
+    delete[] P_;
+    delete[] Q_;
+    a_ = nullptr;
+    indc_ = nullptr;
+    indr_ = nullptr;
+    p_ = nullptr;
```
kswirydo, line 544: doesnt this need to be deleted before becoming a nullptr?

resolve/LinSolverDirectLUSOL.cpp
```diff
@@ -333,20 +536,24 @@ namespace ReSolve
...
+    q_ = nullptr;
```
kswirydo, line 545: same with this, and other variables/

resolve/LinSolverDirectLUSOL.cpp
```diff
@@ -170,7 +192,7 @@ namespace ReSolve
 
   int LinSolverDirectLUSOL::solve(vector_type* rhs, vector_type* x)
   {
-    if (rhs->getSize() != m_ || x->getSize() != n_) {
+    if (rhs->getSize() != m_ || x->getSize() != n_ || !is_factorized_) {
```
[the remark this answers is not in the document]
pelesh: To be tracked in #183.

resolve/LinSolverDirectLUSOL.cpp
pelesh: I believe this loop just adds ones to the diagonal of L.

resolve/LinSolverDirectLUSOL.cpp
pelesh: It does get deleted first.

resolve/LinSolverDirectLUSOL.cpp
superwhiskers: to elaborate, this loop corrects the size of each column segment by the lengths necessary to accommodate an extra element in itself and all previous columns. it then inserts the one at the end of the column's allocated space, since it
1. will always be the last element in a sorted column
2. would add extra computational overhead if it were placed at the beginning (since it would incur an extra n - 1 swaps or something like that) to get the column sorted

sorting of the elements lusol writes to the l factor is done in the next loop, and that is done with a kind of insertion sort, as the comment above it states

pelesh: Thanks for helpful additional explanations, @superwhiskers.

PR #181, opened 2024-08-06
Only build spack base image if necessary

### LINKED ISSUE #171: Suggested modifications to speed up the spack pipelines
A few ideas to speed up the spack pipelines:

1. only re-build base image if it doesn't exist
1. cache a spack.lock so we don't have to concretize

1 is definitely a no brainer, and should be done.

2 is a little more complicated. You would need to remember to remove that file when you update the spack submodule. Additionally we would have to maintain a lock file for each build in the matrix.

_Originally posted by @cameronrutherford in https://github.com/ORNL/ReSolve/issues/170#issuecomment-2182896813_

cameronrutherford:
Closes #171 partially. There is still some work to be done to verify that we need to re-build things with a spack lock / package hash, but that can be made into another issue. We also always force push at the end...

### DIFF
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -7,7 +7,7 @@ env:
   # Our repo name contains upper case characters, so we can't use ${{ github.repository }}
   IMAGE_NAME: resolve
   USERNAME: resolve-bot
-  BASE_VERSION: ubuntu-22.04-fortran
+  BASE_VERSION: ubuntu-22.04-fortran-v0.1.0
 
 # Until we remove the need to clone submodules to build, this should on be in PRs
 on: [pull_request]
@@ -15,7 +15,7 @@ on: [pull_request]
 jobs:
   base_image_build:
     name: Build base image
-    runs-on: ubuntu-22.04-v0.1.0
+    runs-on: ubuntu-22.04
     permissions:
       contents: read
       packages: write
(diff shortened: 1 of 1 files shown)


cameronrutherford:
Haven't tried to re-build the base image, but should be as simple as changing the name of the tag at the top of the workflow YAML. PR works as advertised and speeds up initial pipeline from 1 minute down to 18 seconds ◡̈

cameronrutherford:
Works as advertised. Ready to merge from my perspective, and the re-building of the base image works as well.

.github/workflows/spack_cpu_build.yaml
```diff
@@ -7,19 +7,18 @@ env:
   # Our repo name contains upper case characters, so we can't use ${{ github.repository }}
   IMAGE_NAME: resolve
   USERNAME: resolve-bot
-  BASE_VERSION: ubuntu-22.04-fortran
+  BASE_VERSION: ubuntu-22.04-fortran-v0.1.0
```
pelesh, line 11: Isn't this too specific? Why do we need `-v0.1.0`?

pelesh (approved):
All tests run correctly and pass. I just have one minor question before merging this.

.github/workflows/spack_cpu_build.yaml
cameronrutherford: Whenever we want to rebuild the base container image that our other ghcr packages inherit from, we either need to:
- modify this tag to be unique
- delete the old image and have the pipeline rebuild

If you want to change this tag to something else, be my guest, we just need a way to update/modify moving forward.

PR #182, opened 2024-08-28
remove -1-filling of arrays from lusol factor extraction code

superwhiskers:
this removes -1-filling of arrays from the factor extraction code in #178, making it able to be merged

### DIFF
--- a/resolve/LinSolverDirectLUSOL.cpp
+++ b/resolve/LinSolverDirectLUSOL.cpp
@@ -239,7 +239,6 @@ namespace ReSolve
 
     index_type* columns = L_->getColData(memory::HOST);
     index_type* rows = L_->getRowData(memory::HOST);
-    std::fill_n(rows, current_nnz, -1);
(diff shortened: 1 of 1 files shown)


resolve/LinSolverDirectLUSOL.cpp
```diff
@@ -239,7 +239,6 @@ namespace ReSolve
 
     index_type* columns = L_->getColData(memory::HOST);
     index_type* rows = L_->getRowData(memory::HOST);
-    std::fill_n(rows, current_nnz, -1);
     real_type* values = L_->getValues(memory::HOST);
```
pelesh, line 242: Shouldn't elements of `rows` array be initialized to zero instead?

superwhiskers: That shouldn't be necessary regardless. Some testing I performed on the coo2csr/coo2coo yielded no indication that it would be necessary

pelesh (approved):

PR #184, opened 2024-09-17
Refactor matrix file I/O

### LINKED ISSUE #166: document invariants for `nnz_expanded_` or remove it entirely

### LINKED ISSUE #176: `Sparse::updateData` reads nnz from `nnz_expanded_` if the `is_expanded_` flag is set
when determining size of the buffers to copy data to, `Sparse::updateData` [reads the number of nonzero values within the matrix from the `nnz_expanded_` field if the matrix is marked as expanded](https://github.com/ORNL/ReSolve/blob/develop/resolve/matrix/Csr.cpp#L204), when it should be using `nnz_` in all cases, for if the matrix is expanded, `nnz_` should be set to `nnz_expanded_` anyway. this is confusing behavior, and that branch should be removed in all copies of the `Sparse::updateData` source to remediate it, which should not fail on correctly behaving code. this branch is present in all copies of this method

### LINKED ISSUE #179: `ReSolve::io::writeMatrixToFile` calls `Sparse::print`, which writes out 0-indexed coordinates of matrix data, when the matrix market coordinate format is 1-indexed
the implementation of `ReSolve::io::writeMatrixToFormat` writes the [matrix market](https://math.nist.gov/MatrixMarket/formats.html#MMformat) coordinate data as 0-indexed, when it is specified to be 1-indexed. this makes matrix market files generated by resolve unreadable to any matrix market implementation, [including resolve itself](https://github.com/ORNL/ReSolve/blob/develop/resolve/matrix/io.cpp#L50-L55)

pelesh:
The idea behind this PR is to have matrix import from file utilities in Re::Solve return a ready to use matrix in proper CSR or sorted COO format. That could help simplify some of the code in Re::Solve that was added to support unsorted COO with duplicates to CSR conversions. 

What i hope to accomplish is:
- [x] Import proper CSR from Matrix Market file.
- [x] Remove `nnz_expanded_` sparse matrix member variable and supporting accessor methods (i.e. resolve #166).
- [x] Fix #179.
- [x] Resolve #176.


pelesh:
This PR touches a large number of files but most of the changes are trivial. Significant changes are only in `matrix/io.cpp` file.

pelesh:
All tests pass with NVIDIA and AMD GPUs.

CC: @superwhiskers

resolve/matrix/MatrixHandler.cpp
```diff
@@ -109,16 +108,6 @@ namespace ReSolve {
     }
   }
 
-  /**
```
kswirydo, line 112: While I like the new interface, imagine a situation like this: instead of reading the matrix from a file, we get it from other code (say HiOP) and it comes as COO. In this case, how do we convert it to CSR without going though some loops and hoops?

resolve/matrix/Csr.cpp
```diff
@@ -367,29 +351,19 @@ namespace ReSolve
     } // switch
   }
 
-  int matrix::Csr::updateFromCoo(matrix::Coo* A_coo, memory::MemorySpace memspaceOut)
```
kswirydo, line 370: Again, if we have COO matrix that was NOT read from a file but provided from an application code, how do we update when new data arrives?

resolve/matrix/Csr.cpp
```diff
@@ -28,16 +27,6 @@ namespace ReSolve
   {
   }
 
-  matrix::Csr::Csr(matrix::Coo* A_coo, memory::MemorySpace memspace)
```
kswirydo, line 31: WHy did this constructor dissapear?

resolve/matrix/Coo.hpp
```diff
@@ -13,6 +13,16 @@ namespace ReSolve { namespace matrix {
                 index_type nnz,
                 bool symmetric,
                 bool expanded);
+      Coo(index_type n,
+          index_type m,
+          index_type nnz,
+          bool symmetric,
+          bool expanded,
+          index_type** rows,
```
kswirydo, line 21: were double pointers always here??? I dont remember placing them avywhere in the code.

resolve/matrix/Coo.cpp
```diff
@@ -26,6 +26,101 @@ namespace ReSolve
...
   
+  /**
+   * @brief Hijacking constructor
+   */
+  matrix::Coo::Coo(index_type n,
+                   index_type m,
+                   index_type nnz,
+                   bool symmetric,
+                   bool expanded,
+                   index_type** rows,
```
kswirydo, line 37: Were double pointers always here?

resolve/matrix/io.cpp
```diff
@@ -3,198 +3,522 @@
...
+{ 
+  
+  /**
+   * @brief Helper class for COO matrix sorting.
+   * 
+   * The entire code is in this file. Its scope is to support matrix file I/O
+   * only.
+   * 
+   */
+  class CooTriplet
```
kswirydo, line 25: I would remove `Coo` from the name and kept it as "TripletList" or "InputTripletList" or "RawTripletList"

resolve/matrix/Coo.cpp
pelesh: Yes, double pointers are needed for hijacking. It allows the constructor to set input pointers to NULL, so that the Coo instance has sole control over matrix data.

resolve/matrix/MatrixHandler.cpp
pelesh: This is a good point. Libraries like HiOp and Ipopt have their own COO to CSR conversion utilities and HiOp is developing native CSR matrix support. Each of these packages has different specifications for COO format. HiOp, for example, uses row-major ordered COO without duplicates. I think it is better to handle issues like that on a case-by-case basis.

In this PR, I am suggesting to develop additional utilities only when needed and as a part of Re::Solve interface to specific library rather than having them as a part of the Re::Solve code. For now, Re::Solve will provide only Matrix Market file format import for offline testing.

resolve/matrix/io.cpp
```diff
@@ -3,198 +3,522 @@
...
-      Logger::error() << "Matrix pointer is NULL!\n";
-      return -1;
+    int removeDuplicates(std::list<CooTriplet>& tmp)
+    {
+      std::list<CooTriplet>::iterator it = tmp.begin();
+      while (it != tmp.end())
+      {
+        std::list<CooTriplet>::iterator it_tmp = it;
+        it++;
+        if (*it == *it_tmp) {
```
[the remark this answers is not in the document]
pelesh: We first add duplicate value to the first triplet with the same row and column indices. Only then the list element is erased (meaning removed from the list).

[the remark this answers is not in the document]
pelesh: > Would this work if you have more than 2 duplicates? Like 1 1 13.0 1 1 -10.0 1 1 0.5
> 

Yes, this will work. I just added triplicate in the unit test to confirm it.

resolve/matrix/Csr.cpp
pelesh: The recommendation here is to develop these when needed and optimize them for specifics of the calling application COO format (e.g. no duplicates allowed).

resolve/matrix/Csr.cpp
pelesh: It is not needed anymore. We can easily bring it back, if needed again.

resolve/matrix/Coo.hpp
pelesh: Yes, hijacking constructor needs that.

resolve/matrix/io.cpp
pelesh: Changed to `MatrixElementTriplet`.

resolve/matrix/io.cpp
```diff
@@ -3,198 +3,522 @@
...
+      {}
+      MatrixElementTriplet(index_type i, index_type j, real_type v) : rowidx_(i), colidx_(j), value_(v)
+      {}
+      ~MatrixElementTriplet()
+      {}
+      void setColIdx (index_type new_idx)
+      {
+        colidx_ = new_idx;
+      }
+      void setValue (real_type new_value)
```
maksud, line 38: setColIdx and setValue is there but setRowIdx is missing. Is there any restrictions imposed here?

pelesh: Good catch! This was a dead code. I removed it.

resolve/matrix/MatrixHandler.cpp
kswirydo: okay.

kswirydo (approved):

resolve/matrix/Csr.hpp
```diff
@@ -43,12 +41,10 @@ namespace ReSolve { namespace matrix {
 
       virtual int allocateMatrixData(memory::MemorySpace memspace); 
 
-      virtual void print(std::ostream& file_out = std::cout);
+      virtual void print(std::ostream& file_out = std::cout, index_type indexing_base = 0);
 
       virtual int copyData(memory::MemorySpace memspaceOut);
```
maksud, line 46: Just a very minor comment, copyData sounds like we are providing the data to copy or something. Basically, you are telling the matrix containers to sync data between device and host. So, something that sounds like that may be a good option. For example, syncData.

examples/r_KLU_GLU.cpp
```diff
@@ -82,13 +81,9 @@ int main(int argc, char *argv[])
...
     }
+    bool is_expand_symmetric = true;
     if (i == 0) {
-      A_coo = ReSolve::io::readMatrixFromFile(mat_file);
-      A = new ReSolve::matrix::Csr(A_coo->getNumRows(),
-                                   A_coo->getNumColumns(),
-                                   A_coo->getNnz(),
-                                   A_coo->symmetric(),
-                                   A_coo->expanded());
+      A = ReSolve::io::readCsrMatrixFromFile(mat_file, is_expand_symmetric);
```
maksud, line 86: This example was previously ran with COO matrices. Now, the examples are written with Csr Matrices. Am I correct?

resolve/matrix/io.cpp
```diff
@@ -3,198 +3,750 @@
...
-    if(!file) {
-      Logger::error() << "Empty input to " << __func__ << " function ... \n" << std::endl;
-      return nullptr;
+      index_type m = 0, n = 0, nnz = 0;
+      bool symmetric = false;
+      bool expanded = true;
+
+      std::list<MatrixElementTriplet> tmp;
+
+      readListFromFile(file, is_expand_symmetric, tmp, n, m, nnz, symmetric, expanded);
```
maksud, line 213: By the name of the function readCsrMatrixFromFile I was expecting to read a Csr file not read a COO/MM matrix file. The intention is reading a COO/MM file and convert to CSR. So, readMatrixFromFileAsCsr or something like that may be more meaningful and clear.

examples/r_KLU_GLU.cpp
maksud: Figured out it is still reading COO/MM files and converts to CSR without the end user to explicitly convert it.

maksud:
Functionality wise more concise. Thanks, all test works

resolve/matrix/Csr.hpp
pelesh: This is my pet peeve, too. @kswirydo

pelesh: This is outside the scope of this PR. Will be tracked in #187.

resolve/matrix/io.cpp
pelesh: How about `createCsrFromFile`? This is what this function does actually.

examples/r_KLU_GLU.cpp
pelesh: Yes, with the ability to read matrix as CSR directly from MM file, we avoid the COO to CSR conversion step. I think that makes examples a little bit cleaner as redundant detail is hidden now from the user.

resolve/matrix/io.cpp
pelesh: Addressed in [27dd3f1](https://github.com/ORNL/ReSolve/pull/184/commits/27dd3f1b5ee1b4ee78df4c021889350d1b029f25).

resolve/matrix/io.hpp
```diff
@@ -7,13 +7,16 @@ namespace ReSolve { namespace vector {
...
   class Sparse;
   class Coo;
+  class Csr;
 }}
 
 namespace ReSolve { namespace io {
   using vector_type = vector::Vector;
 
-  matrix::Coo* readMatrixFromFile(std::istream& file);
+  matrix::Coo* readMatrixFromFile(std::istream& file, bool is_expand_symmetric = true);
```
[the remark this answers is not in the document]
pelesh: Addressed in [27dd3f1](https://github.com/ORNL/ReSolve/pull/184/commits/27dd3f1b5ee1b4ee78df4c021889350d1b029f25).

resolve/matrix/io.cpp
pelesh: @kswirydo, please take a look at the name changes I made and chime in, if you have different suggestions.

examples/r_KLU_GLU.cpp
```diff
@@ -98,23 +93,22 @@ int main(int argc, char *argv[])
...
     std::cout<<"Finished reading the matrix and rhs, size: "<<A->getNumRows()<<" x "<<A->getNumColumns()<< ", nnz: "<< A->getNnz()<< ", symmetric? "<<A->symmetric()<< ", Expanded? "<<A->expanded()<<std::endl;
     mat_file.close();
     rhs_file.close();
 
-    //Now convert to CSR.
+    // Update host and device data.
     if (i < 1) {
-      A->updateFromCoo(A_coo, ReSolve::memory::HOST);
       vec_rhs->update(rhs, ReSolve::memory::HOST, ReSolve::memory::HOST);
       vec_rhs->setDataUpdated(ReSolve::memory::HOST);
```
maksud, line 106: This line is not necessary as the update function is taking care of it.

examples/r_KLU_GLU_matrix_values_update.cpp
```diff
@@ -101,29 +96,28 @@ int main(int argc, char *argv[])
...
       std::cout<<"Finished reading the matrix and rhs, size: "<<A->getNumRows()<<" x "<<A->getNumColumns()<< ", nnz: "<< A->getNnz()<< ", symmetric? "<<A->symmetric()<< ", Expanded? "<<A->expanded()<<std::endl;
       mat_file.close();
       rhs_file.close();
 
-      //Now convert to CSR.
+      // Update host and device data.
       if (i < 1) { 
-        A->updateFromCoo(A_coo, ReSolve::memory::HOST);
         vec_rhs->update(rhs, ReSolve::memory::HOST, ReSolve::memory::HOST);
         vec_rhs->setDataUpdated(ReSolve::memory::HOST);
```
maksud, line 115: Not needed, redundant.

maksud (approved):
The interface looks good to me.

examples/r_KLU_GLU_matrix_values_update.cpp
pelesh: See #188.

PR #186, opened 2024-09-19
Update spack_cpu_build.yaml to not checkout in base image build step

cameronrutherford:
In debugging https://github.com/LLNL/hiop/pull/695, I realised that the clone happens at multiple steps in the pipelines, and doesn't need to.

By not cloning when building the base image, we speed up the initial builds a little. We could also kick off the base image pipeline in a separate workflow, and it would surely complete in time for a full spack build (the spack job only needs it to upload to the remote registry.

There's a potential race condition with that if the spack pipeline fails quickly... so I don't see a need to do that.

### DIFF
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -20,13 +20,6 @@ jobs:
       contents: read
       packages: write
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-        with:
-          # Once we move submodule deps into spack, we can do some more builds
-          # Also need to change build script to use spack from base image
-          submodules: true
-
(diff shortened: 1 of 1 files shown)


pelesh (approved):
Everything seems to work fine.

@cameronrutherford, could you please explain why this change? Just to document it for posterity ;).

cameronrutherford:
We already clone the repo in the matrix of jobs in [line 83](https://github.com/ORNL/ReSolve/blob/ab08f211f12e3dc45974fe3b30208b36c5c13bca/.github/workflows/spack_cpu_build.yaml#L105)
of the updated file, and so this is unecessary.

pelesh:
CC @stonecoldhughes

Issue #88, opened 2023-11-29
Consider using smart pointers in ReSolve


pelesh:
Based on @cameronrutherford comment: This is endemic to ExaGO and HiOp, but I would be curious if we could get away from managing raw pointers like this, and have more of our examples use more modern C++ resource management techniques with shared/unique pointers where absolutely necessary, and then just having non-pointer variables here that are cleaned up without `delete`.

_Originally posted by @cameronrutherford in https://github.com/ORNL/ReSolve/pull/83#discussion_r1409384571_

cameronrutherford:
I take this as you agree with my suggestion, but this change will require a little bit of work :)

pelesh:
> I take this as you agree with my suggestion, but this change will require a little bit of work :)

We can do it in incremental steps. For vector/matrix data I think we need to keep raw pointers for portability. But for objects such as Gram-Schmidt, it would make more sense to use smart pointers.

CC @kswirydo

PR #189, opened 2024-09-20
Add functionality test for LUSOL

### LINKED ISSUE #158: Add functionality test for LUSOL
Add functionality test for LUSOL; use functionality test for KLU as a template for the implementation. Note that LUSOL does not have a refactorization option, so just do both solves as in KLU test but with full factorization.

pelesh:
The test is based on KLU functionality test but because LUSOL does not have refactorization option, the test simply performs two full factorizations for different matrices.

This is based on @superwhiskers test in #164.

Resolves #158.

### DIFF
--- a/resolve/LinSolverDirectLUSOL.cpp
+++ b/resolve/LinSolverDirectLUSOL.cpp
@@ -473,9 +473,10 @@ namespace ReSolve
 
   int LinSolverDirectLUSOL::allocateSolverData()
   {
-    // NOTE: determines a hopefully "good enough" size for a_, indc_, indr_.
-    //       see lena_'s documentation for more details
-    lena_ = std::max({2 * nelem_, 10 * m_, 10 * n_, 10000});
+    // LUSOL does not do symbolic analysis to determine workspace size to store
+    // L and U factors, so we have to guess something. See documentation for 
+    // lena_ in resolve/lusol/lusol.f90 file.
+    lena_ = std::max({20 * nelem_, 10 * m_, 10 * n_, 10000});
 
     a_ = new real_type[lena_];
     indc_ = new index_type[lena_];
--- a/tests/functionality/testLUSOL.cpp
+++ b/tests/functionality/testLUSOL.cpp
@@ -30,7 +30,7 @@ using real_type = ReSolve::real_type;
 using vector_type = ReSolve::vector::Vector;
 
 // Prototype for Coo to CSR matrix conversion function
-static int coo2csr_simple(ReSolve::matrix::Coo* A_coo, ReSolve::matrix::Csr* A_csr, ReSolve::memory::MemorySpace memspace);
+static int coo2csr(ReSolve::matrix::Coo* A_coo, ReSolve::matrix::Csr* A_csr, ReSolve::memory::MemorySpace memspace);
 
 int main(int argc, char* argv[])
 {
@@ -106,16 +106,16 @@ int main(int argc, char* argv[])
   vec_diff.update(x_data, ReSolve::memory::HOST, ReSolve::memory::HOST);
 
   ReSolve::matrix::Csr A_csr(A->getNumRows(), A->getNumColumns(), A->getNnz(), A->symmetric(), A->expanded());
-  status = coo2csr_simple(A.get(), &A_csr, ReSolve::memory::HOST);
+  error_sum += coo2csr(A.get(), &A_csr, ReSolve::memory::HOST);
 
   matrix_handler.setValuesChanged(true, ReSolve::memory::HOST);
   error_sum += matrix_handler.matvec(&A_csr,
-                                      &vec_x,
-                                      &vec_r,
-                                      &ONE,
-                                      &MINUSONE,
-                                      "csr",
-                                      ReSolve::memory::HOST);
+                                     &vec_x,
+                                     &vec_r,
+                                     &ONE,
+                                     &MINUSONE,
+                                     "csr",
+                                     ReSolve::memory::HOST);
 
   real_type normRmatrix = sqrt(vector_handler.dot(&vec_r, &vec_r, ReSolve::memory::HOST));
   real_type normXtrue = sqrt(vector_handler.dot(&vec_x, &vec_x, ReSolve::memory::HOST));
@@ -127,24 +127,24 @@ int main(int argc, char* argv[])
   // compute the residual using exact solution
   vec_r.update(rhs, ReSolve::memory::HOST, ReSolve::memory::HOST);
   error_sum += matrix_handler.matvec(&A_csr,
-                                      &vec_test,
-                                      &vec_r,
-                                      &ONE,
-                                      &MINUSONE,
-                                      "csr",
-                                      ReSolve::memory::HOST);
+                                     &vec_test,
+                                     &vec_r,
+                                     &ONE,
+                                     &MINUSONE,
+                                     "csr",
+                                     ReSolve::memory::HOST);
   real_type exactSol_normRmatrix = sqrt(vector_handler.dot(&vec_r, &vec_r, ReSolve::memory::HOST));
 
   // evaluate the residual ON THE CPU using COMPUTED solution
   vec_r.update(rhs, ReSolve::memory::HOST, ReSolve::memory::HOST);
 
   error_sum += matrix_handler.matvec(&A_csr,
-                                      &vec_x,
-                                      &vec_r,
-                                      &ONE,
-                                      &MINUSONE,
-                                      "csr",
-                                      ReSolve::memory::HOST);
+                                     &vec_x,
+                                     &vec_r,
+                                     &ONE,
+                                     &MINUSONE,
+                                     "csr",
+                                     ReSolve::memory::HOST);
 
   real_type normRmatrixCPU = sqrt(vector_handler.dot(&vec_r, &vec_r, ReSolve::memory::HOST));
 
@@ -208,7 +208,7 @@ int main(int argc, char* argv[])
   vec_r.update(rhs, ReSolve::memory::HOST, ReSolve::memory::HOST);
   vec_diff.update(x_data, ReSolve::memory::HOST, ReSolve::memory::HOST);
 
-  status = coo2csr_simple(A.get(), &A_csr, ReSolve::memory::HOST);
+  error_sum += coo2csr(A.get(), &A_csr, ReSolve::memory::HOST);
 
   matrix_handler.setValuesChanged(true, ReSolve::memory::HOST);
   error_sum += matrix_handler.matvec(&A_csr,
@@ -229,24 +229,24 @@ int main(int argc, char* argv[])
   // compute the residual using exact solution
   vec_r.update(rhs, ReSolve::memory::HOST, ReSolve::memory::HOST);
   error_sum += matrix_handler.matvec(&A_csr,
-                                      &vec_test,
-                                      &vec_r,
-                                      &ONE,
-                                      &MINUSONE,
-                                      "csr",
-                                      ReSolve::memory::HOST);
+                                     &vec_test,
+                                     &vec_r,
+                                     &ONE,
+                                     &MINUSONE,
+                                     "csr",
+                                     ReSolve::memory::HOST);
   exactSol_normRmatrix = sqrt(vector_handler.dot(&vec_r, &vec_r, ReSolve::memory::HOST));
 
   // evaluate the residual ON THE CPU using COMPUTED solution
   vec_r.update(rhs, ReSolve::memory::HOST, ReSolve::memory::HOST);
 
   error_sum += matrix_handler.matvec(&A_csr,
-                                      &vec_x,
-                                      &vec_r,
-                                      &ONE,
-                                      &MINUSONE,
-                                      "csr",
-                                      ReSolve::memory::HOST);
+                                     &vec_x,
+                                     &vec_r,
+                                     &ONE,
+                                     &MINUSONE,
+                                     "csr",
+                                     ReSolve::memory::HOST);
 
   normRmatrixCPU = sqrt(vector_handler.dot(&vec_r, &vec_r, ReSolve::memory::HOST));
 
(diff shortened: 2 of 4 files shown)


tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,330 @@
...
+#include <resolve/LinSolverDirectLUSOL.hpp>
+#include <resolve/matrix/Coo.hpp>
+#include <resolve/matrix/Csr.hpp>
+#include <resolve/matrix/MatrixHandler.hpp>
+#include <resolve/matrix/io.hpp>
+#include <resolve/vector/Vector.hpp>
+#include <resolve/vector/VectorHandler.hpp>
+#include <resolve/workspace/LinAlgWorkspace.hpp>
+
+using namespace ReSolve::constants;
```
stonecoldhughes, line 26: Some low-priority nit picking - I think it is slightly better to avoid "using" whenever possible. While it does make the code less verbose, I think it tends to make the code more difficult to understand, since the types in the code are not fully written out where they occur. Again, this is not a high priority suggestion and many disagree.

tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,330 @@
...
+using namespace ReSolve::colors;
+using index_type = ReSolve::index_type;
+using real_type = ReSolve::real_type;
+using vector_type = ReSolve::vector::Vector;
+
+// Prototype for Coo to CSR matrix conversion function
+static int coo2csr(ReSolve::matrix::Coo* A_coo, ReSolve::matrix::Csr* A_csr, ReSolve::memory::MemorySpace memspace);
```
stonecoldhughes, line 33: Is it possible to avoid this forward declaration by including headers differently? Forward declarations are very helpful sometimes, but they introduce a new place for a bug to hide: https://stackoverflow.com/questions/14381348/what-are-the-dangers-of-forward-declarations

tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+  // argv[1] contains the path to the data directory, if it is not ./
+  const std::string data_path = (argc == 2) ? argv[1] : "./";
+
+  ReSolve::LinAlgWorkspaceCpu workspace;
+  ReSolve::MatrixHandler matrix_handler(&workspace);
+  ReSolve::VectorHandler vector_handler(&workspace);
+  ReSolve::LinSolverDirectLUSOL lusol;
+
+  std::string matrix_one_path = data_path + "data/matrix_ACTIVSg200_AC_10.mtx";
```
stonecoldhughes, line 48: if this data file's enclosing directory structure changes, it could crash the program if the developer forgets to update this path. It may be less error prone to configure input files differently using the build system: CMake's https://cmake.org/cmake/help/latest/command/configure_file.html. Internally, a preprocessor variable can be set during build to contain the correct path to the data. These #defined variables can then be included in a header file. The argc/argv feature allows the base path of the data directory to be configured a bit more easily, but argc and argv can be bug prone if there is not lots of logic to ensure well-formed input. It does not appear like such logic is available here.

tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,330 @@
...
+  std::string matrix_two_path = data_path + "data/matrix_ACTIVSg200_AC_11.mtx";
+
+  std::string rhs_one_path = data_path + "data/rhs_ACTIVSg200_AC_10.mtx.ones";
+  std::string rhs_two_path = data_path + "data/rhs_ACTIVSg200_AC_11.mtx.ones";
+
+  std::ifstream matrix_file(matrix_one_path);
+  if (!matrix_file.is_open()) {
```
stonecoldhughes, line 55: now I see that there is some basic checking, this is good

tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+  real_type* x = new real_type[A->getNumRows()];
+  vector_type vec_rhs(A->getNumRows());
+  vector_type vec_x(A->getNumRows());
+  vector_type vec_r(A->getNumRows());
+
+  vec_rhs.update(rhs, ReSolve::memory::HOST, ReSolve::memory::HOST);
+  vec_rhs.setDataUpdated(ReSolve::memory::HOST);
+  vec_x.allocate(ReSolve::memory::HOST);
+
```
stonecoldhughes, line 81: all the code above this line appears to be "setup" code just for creating data structures etc. Perhaps it would improve readability to encapsulate this "setup" functionality in a separate function, and clearly delineate which tasks belong in the "setup" phase

tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,330 @@
...
+  std::cout << "Results: \n";
+  std::cout << "\t ||b-A*x||_2                 : " << std::setprecision(16) << normRmatrix << " (residual norm)\n";
+  std::cout << "\t ||b-A*x||_2  (CPU)          : " << std::setprecision(16) << normRmatrixCPU << " (residual norm)\n";
+  std::cout << "\t ||b-A*x||_2/||b||_2         : " << normRmatrix / normB << " (scaled residual norm)\n";
+  std::cout << "\t ||x-x_true||_2              : " << normDiffMatrix << " (solution error)\n";
+  std::cout << "\t ||x-x_true||_2/||x_true||_2 : " << normDiffMatrix / normXtrue << " (scaled solution error)\n";
+  std::cout << "\t ||b-A*x_exact||_2           : " << exactSol_normRmatrix << " (control; residual norm with exact solution)\n\n";
+
+  delete[] rhs;     // rhs    = nullptr;
```
stonecoldhughes, line 159: these delete statements can be safely removed if smart pointers are used instead of raw pointers: https://en.cppreference.com/w/cpp/memory/unique_ptr

it's always a good idea to use smart pointers because they guarantee that memory is freed once the pointer goes out of scope, effectively insuring against memory leaks.

tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,330 @@
...
+    error_sum += status;
+    return 1;
+  }
+
+  status = lusol.solve(&vec_rhs, &vec_x);
+  if (status != 0) {
+    error_sum += status;
+    return 1;
+  }
+
```
stonecoldhughes, line 97: the code below this line appears to be "post-processing" type functionality where the output is evaluated for correctness. Similar to the "setup" code, could this be separated into a dedicated function to improve delineation of responsibility and improved readability?

tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,330 @@
...
+  std::fill_n(x_data, A->getNumRows(), 1.0);
+
+  vec_test.update(x_data, ReSolve::memory::HOST, ReSolve::memory::HOST);
+  vec_r.update(rhs, ReSolve::memory::HOST, ReSolve::memory::HOST);
+  vec_diff.update(x_data, ReSolve::memory::HOST, ReSolve::memory::HOST);
+
+  ReSolve::matrix::Csr A_csr(A->getNumRows(), A->getNumColumns(), A->getNnz(), A->symmetric(), A->expanded());
+  error_sum += coo2csr(A.get(), &A_csr, ReSolve::memory::HOST);
+
+  matrix_handler.setValuesChanged(true, ReSolve::memory::HOST);
```
stonecoldhughes, line 111: Request equation-style comments in the code section below this line: while it is clear that this code calculates a residual, there is a lot going on and it's not immediately obvious what.

tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,330 @@
...
+  delete[] x_data;  // x_data = nullptr;
+  real_type scaled_residual_norm_one = normRmatrix / normB;
+
+  matrix_file = std::ifstream(matrix_two_path);
+  if (!matrix_file.is_open()) {
+    std::cout << "Failed to open " << matrix_two_path << "\n";
+    return 1;
+  }
+
+  A = std::unique_ptr<ReSolve::matrix::Coo>(ReSolve::io::createCooFromFile(matrix_file));
```
stonecoldhughes, line 170: this is good; "A" will be deallocated automatically, the developer does not have to worry with memory leaks

tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,330 @@
...
+ * @brief 
+ * 
+ * @param A_coo - Input COO matrix without duplicates sorted in row-major order
+ * @param A_csr - Output CSR matrix
+ * @param memspace - memory space in the output matrix where the data is copied
+ * @return int
+ * 
+ * @pre A_coo and A_csr matrices must be of the same size and type.
+ */
+int coo2csr(ReSolve::matrix::Coo* A_coo, ReSolve::matrix::Csr* A_csr, ReSolve::memory::MemorySpace memspace)
```
stonecoldhughes, line 293: This type of conversion is surely useful elsewhere in the code - should we consider putting it in a common utilities header?

tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,330 @@
...
+      is_symmetric != A_csr->symmetric()     ||
+      is_expanded  != A_csr->expanded()) {
+    std::cout << "COO and CSR matrices don't match!\n";
+    return 1;
+  }
+
+  /* const */ index_type* rows_coo = A_coo->getRowData(ReSolve::memory::HOST);
+  /* const */ index_type* cols_coo = A_coo->getColData(ReSolve::memory::HOST);
+  /* const */ real_type*  vals_coo = A_coo->getValues(ReSolve::memory::HOST);
+  index_type* row_csr = new index_type[n + 1];
```
stonecoldhughes, line 314: consider using unique_ptr here as well?

tests/functionality/testLUSOL.cpp
```diff
@@ -0,0 +1,330 @@
...
+  rhs = ReSolve::io::createArrayFromFile(rhs_file);
+  rhs_file.close();
+
+  x = new real_type[A->getNumRows()];
+
+  vec_rhs.update(rhs, ReSolve::memory::HOST, ReSolve::memory::HOST);
+  vec_rhs.setDataUpdated(ReSolve::memory::HOST);
+  vec_x.allocate(ReSolve::memory::HOST);
+
+  error_sum += lusol.setup(A.get());
```
stonecoldhughes, line 188: is this another "setup/analyze/factorize/solve" step different from the first one above in the same file? A bit more commentary might be helpful here, as I'm not sure how this section and what follows fits into the grand scheme.

stonecoldhughes (changes requested):
Overall, this file is structured very clearly. Separating the code into more functions, incorporating data files without hardcoded paths, avoiding manual memory management with new/delete, and adding some more high level comments that provide more insight into exactly what is going on with the validation sections would improve it. I selected the "Submit feedback that must be addressed before merging." option before submitting this review: if you prefer, I can instead select  "Submit general feedback without explicit approval" in future reviews.

resolve/LinSolverDirectLUSOL.cpp
```diff
@@ -473,9 +473,10 @@ namespace ReSolve
...
   {
-    // NOTE: determines a hopefully "good enough" size for a_, indc_, indr_.
-    //       see lena_'s documentation for more details
-    lena_ = std::max({2 * nelem_, 10 * m_, 10 * n_, 10000});
+    // LUSOL does not do symbolic analysis to determine workspace size to store
+    // L and U factors, so we have to guess something. See documentation for 
+    // lena_ in resolve/lusol/lusol.f90 file.
+    lena_ = std::max({20 * nelem_, 10 * m_, 10 * n_, 10000});
 
     a_ = new real_type[lena_];
```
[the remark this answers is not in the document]
pelesh: As I wrote in the comments above there is little we can do about this, because LUSOL does not have symbolic analysis to compute sparsity pattern of L and U factors. L and U factors combined are typically _denser_ than the original matrix A = LU, so we have to allocate workspace that is larger than the storage needed for A. Having that storage 20x the storage for the matrix A is probably an overkill but it will allocate a sufficient space for all but extreme cases. We will be wasting some memory for sure, but this the solver will work correctly.

Mind you LUSOL is a _legacy_ code and does not have all the features more recent linear solvers have. At this point, we don't plan to use LUSOL in applications. It is added to Re::Solve for research purposes.

tests/functionality/testLUSOL.cpp
pelesh: I agree with you that we should avoid `using` when possible. In this particular case, the scope of the `using` is in this file only and it actually helps readability of the code quite a bit. The `namespace constants` contains simple constants such as `one`, `sqrt2`, etc.

tests/functionality/testLUSOL.cpp
pelesh: These are just prototypes for static functions. The alternative option is to _define_ static functions before the calling `main` function. I like better to start the code with `main` but this is just a matter of taste.

tests/functionality/testLUSOL.cpp
pelesh: This is how we handle data for functionality tests and I agree there is a lot of room for improvement there. Please create a separate issue with your suggestions how to handle this better.

tests/functionality/testLUSOL.cpp
pelesh: Can you suggest how to do that in a separate issue? It would be good to have a set of common utility functions for all functionality tests. This is out of scope for this PR, though.

tests/functionality/testLUSOL.cpp
pelesh: Similar as for the comment earlier -- please create an issue with a proposal how to create a set of common utility functions for functionality tests. This PR follows the standard pattern for functionality test that involves a lot of copy-paste-tweak coding, which is not sustainable.

This is out of scope for this PR, though.

tests/functionality/testLUSOL.cpp
pelesh: We could but it is low priority right now. This utility is used for LUSOL only because it has uncommon input requirements. It is unlikely we'll need this elsewhere in the near future.

tests/functionality/testLUSOL.cpp
pelesh: We plan to use raw pointers for matrix/vector data in the future because we move that data frequently between CPU and GPU. In such setting, smart pointers solve only half of the problem at the expense of adding some cumbersome coding to extract raw pointers to pass them to memory management functions. Using smart pointers for Re::Solve classes would be desirable, see #88.

tests/functionality/testLUSOL.cpp
pelesh: See earlier comment and #88.

tests/functionality/testLUSOL.cpp
pelesh: Addressed in [9a50a41](https://github.com/ORNL/ReSolve/pull/189/commits/9a50a414219487dd59a1844d1275aebfefe9151a)

tests/functionality/testLUSOL.cpp
pelesh: Addressed in [9a50a41](https://github.com/ORNL/ReSolve/pull/189/commits/9a50a414219487dd59a1844d1275aebfefe9151a)

tests/functionality/testLUSOL.cpp
stonecoldhughes: since the using directive is only in this file, and for constants which are already pretty self-documenting, this seems like a fine usage of "using"

tests/functionality/testLUSOL.cpp
stonecoldhughes: https://github.com/ORNL/ReSolve/issues/192

tests/functionality/testLUSOL.cpp
stonecoldhughes: https://github.com/ORNL/ReSolve/issues/193

tests/functionality/testLUSOL.cpp
stonecoldhughes: https://github.com/ORNL/ReSolve/issues/193

tests/functionality/testLUSOL.cpp
stonecoldhughes: This is an issue related to the heterogeneous nature of split CPU/GPU memory space. For now, being careful to delete raw pointers is the most straightforward option

tests/functionality/testLUSOL.cpp
stonecoldhughes: https://github.com/ORNL/ReSolve/issues/192

stonecoldhughes (approved):

resolve/LinSolverDirectLUSOL.cpp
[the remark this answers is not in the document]
superwhiskers: it's entirely possible to handle the insufficient memory condition gracefully. while i couldn't find a way to resume lu factorization, lusol will report through the integer parameter array the minimum needed space to continue and you can restart factorization from scratch with a larger lena

PR #190, opened 2024-09-24
Matrix handler now checks in the background if the matrix is CSR.

### LINKED ISSUE #180: `MatrixHandler::matrixInfNorm` should not assume that the input matrix is in the csr format
currently, the `matrixInfNorm` method of `MatrixHandler` and its corresponding implementation classes, `MatrixHandlerCpu` et al. (introduced in #102) incorrectly assume that the input matrix will always be in the csr format. this conflicts with the type signature, which indicates the matrix just has to be of the `Sparse` class.

this is bad for multiple reasons:
- users may incorrectly assume that it permitting the input matrix to be any `Sparse` matrix means it works for any such matrix
- this will silently fail on `Coo` matrices, as all of their rows/columns/values are large enough to accept being indexed that way
  - the error manifests as a performance problem that can be traced back to `MatrixHandler::matrixInfNorm`
  - this also results in incorrect data with a nonobvious source
- this will cause memory access errors when used on `Csc` matrices

this should be resolved by
- adding a disclaimer in the documentation about this (immediately)
- modifying the method by
  - not accepting just any matrix and specializing it to `Csr` or
  - generalizing it to work over different matrix formats

pelesh:
Since only CSR format is supported by the `MatrixHandler` methods for matrix-vector product and matrix norm, these methods will now check if the matrix is CSR and, if not, return an error. When (if) `MatrixHandler` supports other sparse formats, we can add switch to those methods to perform computations specific to the (supported) input matrix  format.

This PR:
- Removes input argument from `matvec` method specifying the matrix format.
- Adds member variable with matrix format ID to `matrix::Sparse` class. Each derived class (`Csr`, `Csc`, etc.) sets this ID at construction. This allows `matvec` and `matrixNorm` to check matrix format "under the hood".
- Resolves #180.
- Updates examples and tests with new `MatrixHandler::matvec` interface.

### DIFF
--- a/resolve/matrix/Coo.cpp
+++ b/resolve/matrix/Coo.cpp
@@ -17,6 +17,7 @@ namespace ReSolve
 
   matrix::Coo::Coo(index_type n, index_type m, index_type nnz) : Sparse(n, m, nnz)
   {
+    sparse_format_ = TRIPLET;
   }
   
   matrix::Coo::Coo(index_type n, 
@@ -43,7 +44,6 @@ namespace ReSolve
                    memory::MemorySpace memspaceDst)
     : Sparse(n, m, nnz, symmetric, expanded)
   {
-    using namespace memory;
     sparse_format_ = TRIPLET;
 
     int control = -1;
--- a/resolve/matrix/Coo.cpp
+++ b/resolve/matrix/Coo.cpp
@@ -12,6 +12,7 @@ namespace ReSolve
 
   matrix::Coo::Coo()
   {
+    sparse_format_ = TRIPLET;
   }
 
   matrix::Coo::Coo(index_type n, index_type m, index_type nnz) : Sparse(n, m, nnz)
@@ -24,6 +25,7 @@ namespace ReSolve
                    bool symmetric,
                    bool expanded) : Sparse(n, m, nnz, symmetric, expanded)
   {
+    sparse_format_ = TRIPLET;
   }
   
   /**
@@ -42,6 +44,8 @@ namespace ReSolve
     : Sparse(n, m, nnz, symmetric, expanded)
   {
     using namespace memory;
+    sparse_format_ = TRIPLET;
+
     int control = -1;
     if ((memspaceSrc == memory::HOST)   && (memspaceDst == memory::HOST))  { control = 0;}
     if ((memspaceSrc == memory::HOST)   && (memspaceDst == memory::DEVICE)){ control = 1;}
--- a/resolve/matrix/MatrixHandlerCpu.cpp
+++ b/resolve/matrix/MatrixHandlerCpu.cpp
@@ -33,6 +33,20 @@ namespace ReSolve {
 
   /**
    * @brief result := alpha * A * x + beta * result
+   * 
+   * @param[in]     A - matrix
+   * @param[in]     vec_x - vector multiplied by A
+   * @param[in,out] vec_result - resulting vector
+   * @param[in]     alpha - matrix-vector multiplication factor
+   * @param[in]     beta - sum into result factor
+   * @return int    error code, 0 if successful
+   * 
+   * @pre Matrix `A` is in CSR format.
+   * 
+   * @note If we decide to implement this function for different matrix
+   * format, the check for CSR matrix will be replaced with a switch
+   * statement to select implementation for recognized input matrix
+   * format.
    */
   int MatrixHandlerCpu::matvec(matrix::Sparse* A, 
                                vector_type* vec_x, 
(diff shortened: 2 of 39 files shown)


resolve/matrix/Sparse.hpp
```diff
@@ -8,15 +8,25 @@
...
+   * all nonzero matrix values need to be stored, so all utilities needed for
+   * the values are implemented in this class.
    *
    * @author Kasia Swirydowicz <kasia.swirydowicz@pnnl.gov>
    */
   class Sparse 
   {
+    public:
+      /// Supported sparse matrix formats
+      enum SparseFormat{TRIPLET, COMPRESSED_SPARSE_ROW, COMPRESSED_SPARSE_COLUMN};
```
stonecoldhughes, line 28: a scoped enumeration instead of this unscoped enumeration would provide more rigorous type checking: that may or may not be desired. Adds a little bit more rigor at the expense of a little more verbosity.

resolve/matrix/MatrixHandlerCpu.cpp
```diff
@@ -34,53 +34,55 @@ namespace ReSolve {
...
-        sum *= (*alpha);
-        result_data[i] = result_data[i]*(*beta) + sum;
-      } 
-      vec_result->setDataUpdated(memory::HOST);
-      return 0;
-    } else {
-      out::error() << "MatVec not implemented (yet) for " 
-                   << matrixFormat << " matrix format." << std::endl;
+
+    if (A->getSparseFormat() != matrix::Sparse::COMPRESSED_SPARSE_ROW) {
```
stonecoldhughes, line 45: an "assert" statement could replace conditionals like these if there is ever a desire to disable/enable the checking; while turning off checks is unsafe, there may be performance benefits if these conditionals end up being called a lot. The top answer on this post helps shed more light on the two avenues: https://stackoverflow.com/questions/11468899/any-benefit-of-using-assert-instead-of-using-a-simple-if

stonecoldhughes (changes requested):
Removing the "csr" parameter is a really good improvement here. Minor comments include weighing the benefits of static assertions versus conditionals for checking ; using the "using" directive; extending a variable name to be more descriptive

resolve/matrix/MatrixHandlerCpu.cpp
pelesh: This check will be replaced with a switch statement if/when we implement methods that work with formats other than CSR. I added a note to that end in method comments.

In the meantime, I replaced `if` with `assert` as you suggested.

resolve/matrix/Sparse.hpp
pelesh: Please create a separate issue to discuss this. I believe this will be applicable to other `enum`s in the code.

resolve/matrix/Coo.cpp
```diff
@@ -42,6 +44,8 @@ namespace ReSolve
     : Sparse(n, m, nnz, symmetric, expanded)
   {
     using namespace memory;
+    sparse_format_ = TRIPLET;
+
     int control = -1;
     if ((memspaceSrc == memory::HOST)   && (memspaceDst == memory::HOST))  { control = 0;}
```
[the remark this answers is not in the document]
pelesh: Good catch!

resolve/matrix/Coo.cpp
```diff
@@ -42,6 +44,8 @@ namespace ReSolve
     : Sparse(n, m, nnz, symmetric, expanded)
   {
     using namespace memory;
```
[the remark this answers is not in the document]
pelesh: Removed!

resolve/matrix/Coo.cpp
```diff
@@ -42,6 +44,8 @@ namespace ReSolve
...
+    sparse_format_ = TRIPLET;
```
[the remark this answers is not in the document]
pelesh: Triplet and COO are synonyms. I used "triplet" in this case to prevent possible confusion with class name `Coo`. I would leave it as is.

resolve/matrix/MatrixHandlerCpu.cpp
pelesh: See [67915e1](https://github.com/ORNL/ReSolve/pull/190/commits/67915e10f2a084e418c2fb2739037be6e370879f)

resolve/matrix/Coo.cpp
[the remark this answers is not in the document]
pelesh: See [67915e1](https://github.com/ORNL/ReSolve/pull/190/commits/67915e10f2a084e418c2fb2739037be6e370879f)

resolve/matrix/Coo.cpp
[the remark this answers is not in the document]
stonecoldhughes: Sounds like "TRIPLET" is standard terminology in the field, so this name is fine as it is

resolve/matrix/Sparse.hpp
stonecoldhughes: https://github.com/ORNL/ReSolve/issues/191

stonecoldhughes (approved):

PR #198, opened 2024-10-01
Fix memory leak in `GramSchmidt` class; setup now in GMRES

### LINKED ISSUE #196: Memory leak in `GramSchmidt` class
In reorthogonalization methods Modified 2-Sync Gramm-Schmidt and Postmodern Gramm_Schmidt, member variable `h_rv_` (a pointer to `real_type`) is first allocated
```c++
  h_rv_ = new real_type[num_vecs_ + 1];
``` 
and then, a few lines later, a new value is assigned to the pointer `h_rv_` as
```c++
  h_rv_ = vec_rv_->getVectorData(1, memory::HOST);
```
thus leaking previous allocation.

There is no need for `h_rv_` to be a member variable to begin with, since it is used only locally in Modified 2-sync and Postmodern Gram-Schmidt methods. It should be defined locally in the `orthogonalize` function and initialized to `nullptr`.

CC @stonecoldhughes @maksud

### LINKED ISSUE #197: Who has responsibility to set up `GramSchmidt` class instance?
A pointer to `GramSchmidt` class instance needs to be passed to each (F)GMRES solver instance. This design abstracts away Gram-Schmidt method from the iterative solver implementation and allows for adding new reorthogonalization methods in the future.

However, in current design (and examples), the application calling the (F)GMRES solver has the responsibility to set up the `GramSchmidt` class. The `setup` function takes in two parameters -- the size of the linear system and the number of iterations till solver restart. Having the calling application set these two parameters creates a risk that `GramSchmidt` and iterative solver class can receive different values for the two parameters, possibly leading to hard to detect bugs in the application code.

Consider shifting responsibility of setting up `GramSchmidt` class instance from the calling application to the iterative solver itself. A `setup` method in a (F)GMRES class could set up the `GramSchmidt` instance, as well, since it has access to all needed parameter values for that. Such implementation would ensure the same parameter values are used for (F)GMRES and the `GramSchmidt`class.

CC @stonecoldhughes @maksud

pelesh:
The PR addresses these issues issues:
1. Fix memory leak in `GramSchmidt` class (resolves #196).
2. Make iterative solver owning `GramSchmidt` instance responsible for setting it up (resolves #197).
3. Removes unused FGMRES constructors.
4. Sets flags correctly upon completing `GramSchmidt` instance setup.

### DIFF
--- a/resolve/GramSchmidt.cpp
+++ b/resolve/GramSchmidt.cpp
@@ -164,9 +164,9 @@ namespace ReSolve
           vector_handler_->scal(&t, vec_w_, memspace_);  
         } else {
           assert(0 && "Gram-Schmidt failed, vector with ZERO norm\n");
-          return -1;
+          return 1;
         }
-        break;
+        return 0;
 
       case cgs2:
         vec_v_->setData(V->getVectorData(i + 1, memspace_), memspace_);
@@ -210,10 +210,9 @@ namespace ReSolve
           vector_handler_->scal(&t, vec_v_, memspace_);  
         } else {
           assert(0 && "Gram-Schmidt failed, vector with ZERO norm\n");
-          return -1;
+          return 1;
         }
         return 0;
-        break;
 
       case mgs_two_sync:
         // V[1:i]^T[V[i] w]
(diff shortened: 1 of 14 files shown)


resolve/GramSchmidt.cpp
```diff
@@ -100,32 +100,39 @@ namespace ReSolve
     num_vecs_ = restart;
     if((variant_ == mgs_two_sync) || (variant_ == mgs_pm)) {
       h_L_  = new real_type[num_vecs_ * (num_vecs_ + 1)]();
```
stonecoldhughes, line 102: could this variable be handled similarly to h_rv?

stonecoldhughes:
all tests pass for me locally!

stonecoldhughes (approved):
approved!

Issue #188, opened 2024-09-20
Review examples and remove redundant code


pelesh:
Examples should provide users an idea how to use Re::Solve effectively. Development of some of the examples have not followed Re::Solve core development quite well, so there is some redundant code there that might be misleading users.

It would be a good idea to review and update examples to be more educational of how to use Re::Solve effectively.

_Originally posted by @maksud in https://github.com/ORNL/ReSolve/pull/184#discussion_r1768864945_

pelesh:
Fixed in #199.

PR #199, opened 2024-10-09
Rename `copyData` -> `syncData` and ensure correct flags set when syncing

### LINKED ISSUE #126: When only host is updated, the device is out of date (and vice versa)
In `update` methods in vector and matrix classes it is probably safer to set both update flags at the same time - i.e if cpu data gets updated, gpu data automatically should switch to NOT updated.

_Originally posted by @kswirydo in https://github.com/ORNL/ReSolve/pull/124#discussion_r1429345023_

### LINKED ISSUE #187: Consider renaming matrix::Sparse::copyData to matrix::Sparse::SyncData
Just a comment: `copyData` sounds like we are providing the data to copy or something. Basically, you are telling the matrix containers to sync data between device and host. So, something that sounds like that may be a good option. For example, `syncData`.

_Originally posted by @maksud in https://github.com/ORNL/ReSolve/pull/184#discussion_r1768275445_

pelesh:
There are two issues addressed here:
- Make wholesale name change `copyData` -> `syncData`. Methods `syncData` synchronize data between the host and device within the same vector or matrix object. "Copy" often suggests data is copied from/to another object. Resolves #187. 
- The "update" methods copy matrix or vector data to/from external sources and set flags indicating which memory space within the object is up-to-date. For example, if the host memory is updated, host flag `h_data_updated_` is set to `true` and device flag `d_data_updated_` is set to `false`. A sync method needs to be used if one wants to update both. Resolves #126.

Both, sync and update methods need to check if the operations are meaningful. For example if syncing device memory space, the host needs to be up-to-date.

### DIFF
--- a/examples/r_KLU_GLU.cpp
+++ b/examples/r_KLU_GLU.cpp
@@ -105,7 +105,7 @@ int main(int argc, char *argv[])
       vec_rhs->update(rhs, ReSolve::memory::HOST, ReSolve::memory::HOST);
       vec_rhs->setDataUpdated(ReSolve::memory::HOST);
     } else { 
-      A->copyData(ReSolve::memory::DEVICE);
+      A->syncData(ReSolve::memory::DEVICE);
       vec_rhs->update(rhs, ReSolve::memory::HOST, ReSolve::memory::DEVICE);
     }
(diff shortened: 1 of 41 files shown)


kswirydo (approved):

pelesh:
Resolves #188.

PR #206, opened 2024-12-18
Replace FindKLU with FindSuiteSparse

pelesh:
Replaced `FindKLU.cmake` file with a more general `FindSuiteSparse.cmake`. The new finding function now looks for AMD, COLAMD and KLU libraries. This will be needed for HyKKT integration (currently in works by @shakedregev) as HyKKT links directly to AMD but not KLU. The new finding function can be easily extended to add other libraries from SuiteSparse (I suspect we'll need CHOLMOD soon).

Also changed in this PR -- all external dependencies are off by default. User needs to enable them explicitly now. 

This PR is mainly reusing @cameronrutherford's work in [GridKit](https://github.com/ORNL/GridKit).


shakedregev:
No comments on the code, I ran it successfully.

shakedregev (commented):
No comments on the code, I ran it successfully.

shakedregev (approved):

cmake/FindSuiteSparse.cmake
```diff
@@ -0,0 +1,80 @@
+#[[
+
+Finds Sutiesparse include directory and libraries and exports target `Suitesparse`
+
+User may set:
+- SUITESPARSE_ROOT_DIR
+
+Author(s):
+- Cameron Rutherford <cameron.rutherford@pnnl.gov>
```
cameronrutherford, line 9: I guess this was just copy + pasted ;)

Curious why suitesparse doesn't support `find_package`....

pelesh: That is a good point. I didn't see SuiteSparse packages providing CMake config file even though they use CMake for the build now.

pelesh:
> I just now noticed that `make test_install` actually fails
> 

I can't reproduce this issue and installed `ReSolveTargets.cmake` looks correct to me. @shakedregev, could you please create an issue and post error output you are getting and the content of the `ReSolveTargets.cmake` you get installed?

PR #207, opened 2024-12-20
Fix warnings in LUSOL solver and unit test

pelesh:
Fix warnings when converting from integer index type to `size_t`.


maksud (approved):

superwhiskers (commented):
this seems fine to me

PR #208, opened 2024-12-27
Change member functions and data names in vector and matrix classes to be more consistent

### LINKED ISSUE #167: Use consistent naming scheme across matrix and vector classes
as was brought up in #164 by @pelesh, the current names of these fields is unclear. they should be renamed to something like `owns_gpu_sparsity_pattern_` and `owns_gpu_nonzeroes_` (with equivalents for the cpu)

- rename `owns_gpu_data_`, `owns_gpu_values_`, `owns_cpu_data_`, `owns_cpu_values_` to the proposed names
- rename the methods for setting them on `Sparse`
- add methods for getting them?

pelesh:
Change names of methods and data members in vector and matrix classes to be more consistent and accurate. Specific changes include:
- Matrix classes (`Sparse`, `Csr`, `Csc`, `Coo`)
    - Rename `setMatrixData` to `setDataPointers`
    - Rename `setNewValues` to `setValuesPointers`
    - Rename `updateData` to `copyData`
    - Change input data arrays in `copyData` to `const` arrays.
    - Rename `owns_*_data_` to `owns_*_sparsity_pattern_`
- Vector class
    - Rename `deepCopyVectorData` to `copyDataTo`
    - Rename `update` to `copyDataFrom`.

Fixes #167

### DIFF
--- a/resolve/matrix/Coo.cpp
+++ b/resolve/matrix/Coo.cpp
@@ -61,8 +61,8 @@ namespace ReSolve
         h_col_data_ = *cols;
         h_val_data_ = *vals;
         h_data_updated_ = true;
-        owns_cpu_vals_ = true;
-        owns_cpu_data_  = true;
+        owns_cpu_values_ = true;
+        owns_cpu_sparsity_pattern_ = true;
         // Make sure there is no device data.
         if (d_row_data_ || d_col_data_ || d_val_data_) {
           out::error() << "Device data unexpectedly allocated. "
@@ -79,8 +79,8 @@ namespace ReSolve
         d_col_data_ = *cols;
         d_val_data_ = *vals;
         d_data_updated_ = true;
-        owns_gpu_vals_ = true;
-        owns_gpu_data_  = true;
+        owns_gpu_values_ = true;
+        owns_gpu_sparsity_pattern_ = true;
         syncData(memspaceDst);
         // Hijack data from the source
         *rows = nullptr;
@@ -93,8 +93,8 @@ namespace ReSolve
         h_col_data_ = *cols;
         h_val_data_ = *vals;
         h_data_updated_ = true;
-        owns_cpu_vals_ = true;
-        owns_cpu_data_  = true;
+        owns_cpu_values_ = true;
+        owns_cpu_sparsity_pattern_ = true;
         syncData(memspaceDst);
         // Hijack data from the source
         *rows = nullptr;
@@ -107,8 +107,8 @@ namespace ReSolve
         d_col_data_ = *cols;
         d_val_data_ = *vals;
         d_data_updated_ = true;
-        owns_gpu_vals_ = true;
-        owns_gpu_data_  = true;
+        owns_gpu_values_ = true;
+        owns_gpu_sparsity_pattern_ = true;
         // Make sure there is no device data.
         if (h_row_data_ || h_col_data_ || h_val_data_) {
           out::error() << "Host data unexpectedly allocated. "
@@ -196,11 +196,11 @@ namespace ReSolve
       if ((h_row_data_ == nullptr) && (h_col_data_ == nullptr)) {
         this->h_row_data_ = new index_type[nnz_current];
         this->h_col_data_ = new index_type[nnz_current];
-        owns_cpu_data_ = true;
+        owns_cpu_sparsity_pattern_ = true;
       }
       if (h_val_data_ == nullptr) {
         this->h_val_data_ = new real_type[nnz_current];
-        owns_cpu_vals_ = true;
+        owns_cpu_values_ = true;
       }
     }
 
@@ -212,11 +212,11 @@ namespace ReSolve
       if ((d_row_data_ == nullptr) && (d_col_data_ == nullptr)) {
         mem_.allocateArrayOnDevice(&d_row_data_, nnz_current);
         mem_.allocateArrayOnDevice(&d_col_data_, nnz_current);
-        owns_gpu_data_ = true;
+        owns_gpu_sparsity_pattern_ = true;
       }
       if (d_val_data_ == nullptr) {
         mem_.allocateArrayOnDevice(&d_val_data_, nnz_current);
-        owns_gpu_vals_ = true;
+        owns_gpu_values_ = true;
       }
     }
 
@@ -275,8 +275,8 @@ namespace ReSolve
       std::fill(h_col_data_, h_col_data_ + nnz_current, 0);  
       this->h_val_data_ = new real_type[nnz_current];
       std::fill(h_val_data_, h_val_data_ + nnz_current, 0.0);  
-      owns_cpu_data_ = true;
-      owns_cpu_vals_ = true;
+      owns_cpu_sparsity_pattern_ = true;
+      owns_cpu_values_ = true;
       return 0;
     }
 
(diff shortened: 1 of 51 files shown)


pelesh:
Rebased to `develop`.

resolve/matrix/Coo.cpp
```diff
@@ -191,32 +191,32 @@ namespace ReSolve
...
-        owns_cpu_vals_ = true;
+        owns_cpu_values_ = true;
       }
     }
 
     if (memspaceOut == memory::DEVICE) {
       //check if cuda data allocated
       if ((d_row_data_ == nullptr) != (d_col_data_ == nullptr)) {
-        out::error() << "In Coo::updateData one of device row or column data is null!\n";
+        out::error() << "In Coo::copyData one of device row or column data is null!\n";
```
shakedregev, line 210: nitpicking: wouldn't it be better to know which one is null? Of course if everything is working they should both be null or neither be null.

resolve/matrix/Csc.hpp
```diff
@@ -19,17 +19,17 @@ namespace ReSolve { namespace matrix {
...
+                           const index_type* col_data,
+                           const real_type* val_data,
+                           memory::MemorySpace memspaceIn,
+                           memory::MemorySpace memspaceOut); 
+      virtual int copyData(const index_type* row_data,
+                           const index_type* col_data,
+                           const real_type* val_data,
+                           index_type new_nnz,
+                           memory::MemorySpace memspaceIn,
+                           memory::MemorySpace memspaceOut);
```
shakedregev, line 32: Is there a reason why we have one creator with `new_nnz` and one without? If anything, wouldn't it make sense to also know `n`?

resolve/vector/Vector.cpp
```diff
@@ -503,25 +503,30 @@ namespace ReSolve { namespace vector {
...
    * @param[in] i          - Index of a vector in a multivector
    * @param[in] memspace   - Memory space (HOST or DEVICE) to copy from
    *
    * @return 0 if successful, -1 otherwise.
    *
-   * @pre _i_ < _k_ i.e,, _i_ is smaller than the total number of vectors in multivector.
-   * @pre _dest_ is allocated, and the size of _dest_ is at least _n_ (lenght of a single vector in the multivector).
+   * @pre _i_ < _k_ i.e,, _i_ is smaller than the total number of vectors
+   * in multivector.
+   * @pre _dest_ is allocated, and the size of _dest_ is at least _n_ (lenght
```
shakedregev, line 517: length mispelled

shakedregev (approved):
Tests are passing, looks good besides the minor comments

nkoukpaizan:
I haven't looked at codes changes, but tests are passing for me as well (tested on Frontier).

resolve/vector/Vector.cpp
```diff
@@ -503,25 +503,30 @@ namespace ReSolve { namespace vector {
   }
 
   /** 
-   * @brief copy HOST or DEVICE data of a specified vector in a multivector to _dest_. 
+   * @brief copy HOST or DEVICE data of a specified vector in a multivector
+   * to _dest_. 
    * 
    * @param[out] dest      - Pointer to the memory to which data is copied
    * @param[in] i          - Index of a vector in a multivector
    * @param[in] memspace   - Memory space (HOST or DEVICE) to copy from
```
nkoukpaizan, line 511: memspace --> memspaceInOut?

nkoukpaizan (approved):

resolve/matrix/Csc.hpp
kswirydo: yes there is a reason. If nnz changes, you need to reallocate. If nnz does not change, we do not.

tests/unit/matrix/LUSOLTests.hpp
```diff
@@ -155,11 +155,11 @@ namespace ReSolve
           // NOTE: these are hardcoded for now
           index_type size = static_cast<index_type>(valsA_.size());
           matrix::Coo* A = new matrix::Coo(9, 9, size, true, true);
-          A->updateData(rowsA_.data(),
-                        colsA_.data(),
-                        valsA_.data(),
-                        memory::HOST,
-                        memory::HOST);
+          A->copyData(rowsA_.data(),
```
kswirydo, line 158: shouldnt this be copyDataFrom ?

tests/unit/matrix/MatrixFactorizationTests.hpp
```diff
@@ -162,7 +162,7 @@ class MatrixFactorizationTests : TestBase
     // Allocate NxN CSR matrix with NNZ nonzeros
     matrix::Csr* A = new matrix::Csr(N, N, NNZ);
     A->allocateMatrixData(memory::HOST);
-    A->updateData(&rowsA_[0], &colsA_[0], &valsA_[0], memory::HOST, memory::HOST);
+    A->copyData(&rowsA_[0], &colsA_[0], &valsA_[0], memory::HOST, memory::HOST);
```
kswirydo, line 165: same here

kswirydo (approved):
My main concern is that now we have `copyData[To/From]` instead of `update` but at the same time, we kept `setDataUpdated` which makes things slightly more confusing.

pelesh:
> My main concern is that now we have `copyData[To/From]` instead of `update` but at the same time, we kept `setDataUpdated` which makes things slightly more confusing.

I am not sure if this makes things more confusing. The purpose of `setDataUpdated` is to tell matrix/vector object that its data was changed externally by accessing its raw data pointers. The `setDataUpdated` is not used in conjunction with `copyData*` functions because the latter set internal matrix/vector flags to indicate if any object data was changed (updated) during the copy. Perhaps a better name would be `setDataChanged`?

resolve/matrix/Coo.cpp
pelesh: I believe your question contains the answer :). If this condition holds, something very wrong is happening in the code and more debugging is needed. Knowing which one is null would be a small but insufficient help. Probably a better way to go about this would be to
```c++
  assert((d_row_data_ == nullptr) == (d_col_data_ == nullptr))
```

resolve/vector/Vector.cpp
pelesh: Addressed in #210.

resolve/matrix/Coo.cpp
shakedregev: Also, this is quite a nifty way to write a not xor, which I think I missed the first read. Maybe this is standard and I just haven't seen it, but an explanation could be good.

resolve/vector/Vector.cpp
pelesh: Addressed in [09b16e8](https://github.com/ORNL/ReSolve/pull/208/commits/09b16e86f0c12570edd039eb8f2b3266aee8aeea).

tests/unit/matrix/MatrixFactorizationTests.hpp
pelesh: Addressed in [0e37fd0](https://github.com/ORNL/ReSolve/pull/208/commits/0e37fd092697145742bb2b3563f4c4c24c07378f).

tests/unit/matrix/LUSOLTests.hpp
pelesh: Addressed in [0e37fd0](https://github.com/ORNL/ReSolve/pull/208/commits/0e37fd092697145742bb2b3563f4c4c24c07378f).

PR #211, opened 2025-01-08
Fix issue with linking workspace to HyKKT.

pelesh:
Make sure a workspace library is linked to HyKKT module in Re::Solve.

Also, if CUDA or HIP are enabled, they can be linked with HyKKT module. The module may not use any of those backends, but it can be linked to them.


maksud (approved):

shakedregev:
Checked on Frontier and on CUDA machine. All tests pass

Issue #183, opened 2024-09-12
::solve method in LUSOL should try to factorize matrix if it was not factorized already


pelesh:
@kswirydo says: I think if matrix exists (and m_\neq 0), it should throw a warning and try to factorize, not  return -1.

_Originally posted by @kswirydo in https://github.com/ORNL/ReSolve/pull/178#discussion_r1757295680_

superwhiskers:
i think we can close this issue now @pelesh

pelesh:
Addressed in #212.

PR #212, opened 2025-01-09
address issue #183

superwhiskers:
this probably addresses all of issue #183.

some additional considerations:

- do we also call the workspace allocation method too (`LinSolverDirectLUSOL::analyze()`)
- are there any other invariants we should check for before automatically factorizing?

### DIFF
--- a/resolve/LinSolverDirectLUSOL.cpp
+++ b/resolve/LinSolverDirectLUSOL.cpp
@@ -146,10 +146,17 @@ namespace ReSolve
 
   int LinSolverDirectLUSOL::factorize()
   {
-    // NOTE: this is probably good enough as far as checking goes
-    if (a_ == nullptr || indc_ == nullptr || indr_ == nullptr) {
-      out::error() << "LUSOL workspace not allocated!\n";
-      return -1;
+    if (!is_solver_data_allocated_) {
+      out::warning() << "LinSolverDirect::factorize() called on "
+                     << "LinSolverDirectLUSOL without allocating the "
+                     << "workspace first!\n";
+
+      // it isn't possible for this to error in any recoverable way
+      // but we'll check anyway
+      int inform = analyze();
+      if (inform < 0) {
+        return inform;
+      }
     }
 
     index_type inform = 0;
@@ -180,7 +187,7 @@ namespace ReSolve
 
     // TODO: consider handling inform = 7
 
-    return inform;
+    return static_cast<int>(inform);
   }
 
   int LinSolverDirectLUSOL::refactorize()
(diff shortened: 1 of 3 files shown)


resolve/LinSolverDirectLUSOL.cpp
```diff
@@ -480,7 +495,7 @@ namespace ReSolve
   int LinSolverDirectLUSOL::allocateSolverData()
   {
     // LUSOL does not do symbolic analysis to determine workspace size to store
-    // L and U factors, so we have to guess something. See documentation for 
+    // L and U factors, so we have to guess something. See documentation for
     // lena_ in resolve/lusol/lusol.f90 file.
     lena_ = std::max({20 * nelem_, 10 * m_, 10 * n_, 10000});
```
kswirydo, line 500: I would suggest making this "magic" numbers constants that are stored in h file (?)

tests/unit/matrix/runLUSOLTests.cpp
```diff
@@ -11,6 +11,7 @@ int main(int, char**)
     ReSolve::tests::LUSOLTests test;
 
     result += test.lusolConstructor();
+    result += test.automaticFactorization();
```
kswirydo, line 14: Don't we also  need to check correctness of the solution or is these done in another test?

resolve/LinSolverDirectLUSOL.cpp
superwhiskers: there isn't much else to explain about these and adding constants would be excessive, in my opinion. they're [taken straight from lusol's source as referenced in the comment](https://github.com/ORNL/ReSolve/blob/develop/resolve/lusol/lusol.f90#L256-L263) save from the `2 * nelem_` being altered to `20 * nelem_` by @pelesh [in the functionality test pull request](https://github.com/ORNL/ReSolve/pull/189/files#r1775611754)

tests/unit/matrix/runLUSOLTests.cpp
superwhiskers: is [this](https://github.com/ORNL/ReSolve/blob/57c62b92a2d3fe58050e79cc6630d123740c8fc1/tests/unit/matrix/LUSOLTests.hpp#L69) not enough?

pelesh:
> * do we also call the workspace allocation method too (`LinSolverDirectLUSOL::analyze()`)

I think it would be good to check if the workspace has been allocated and run `analyse` if not.

tests/unit/matrix/runLUSOLTests.cpp
pelesh: To be consistent, we could rename this test to something like `automaticFactorizationSolve` to avoid confusion.

superwhiskers:
it would probably be better to just edit the factorization method since it errors out if the workspace is not allocated yet and it'd be more consistent for both of them to automatically allocate the workspace if it is not yet already allocated

tests/unit/matrix/runLUSOLTests.cpp
pelesh: Addressed in [70f2f88](https://github.com/ORNL/ReSolve/pull/212/commits/70f2f88576ebd396aca4ae9c63c5c56601552ffc).

pelesh:
> it would probably be better to just edit the factorization method since it errors out if the workspace is not allocated yet and it'd be more consistent for both of them to automatically allocate the workspace if it is not yet already allocated

Please go ahead and do that.

pelesh (approved):
lgtm

PR #222, opened 2025-02-06
Fixed naming inconsistencies

shakedregev:


tests/functionality/testSysGmres.cpp
```diff
@@ -213,7 +213,7 @@ void processInputs(std::string& method, std::string& gs, std::string& sketch)
   if (gs != "cgs1" && 
       gs != "cgs2" && 
       gs != "mgs" && 
-      gs != "mgs_two_sync" 
+      gs != "mgs2_sync" 
       && gs != "mgs_pm") {
```
pelesh, line 217: Actually `mgs_two_sync` is better name than `msg2_sync`, but I see how it can be confused because we have `cgs1` and `cgs2`. I think it is documented somewhere in the code, but for the sake of clarity here:
- CGS1: classical Gram-Schmidt
- CGS2: improved variant of classical Gram-Schmidt
- MGS: modified Gram-Schmidt
- MGS_TWO_SYNC: modified Gram-Schmidt with two synchronization communications

@kswirydo can provide better description of these methods, preferably in Re::Solve docs :).

pelesh (changes requested):
Thanks for doing this. I would just revert back to `mgs_two_sync`.

shakedregev:
> Thanks for doing this. I would just revert back to `mgs_two_sync`.

Reverted @pelesh

PR #225, opened 2025-02-13
Declutter examples

### LINKED ISSUE #221: The examples built do not expect a uniform file name
Some examples take the literal string passed as the file name, I think this is desirable, because it means no need to recompile to change data formats. The script must concatenate the filenames in the correct way.

Other examples internally create the file name by appending the numbers and the suffix `.mtx`. This is not desirable as it requires the data to have a specific suffix and format.

@pelesh @nkoukpaizan @maksud

### LINKED ISSUE #235: Gram-Schmidt method allocated incorrectly in `LinSolverIterativeRandFGMRES` class
In the `LinSolverIterativeRandFGMRES:setup` method, Gram-Schmidt orthogonalization is allocated as
```c++
    GS_->setup(n_, restart_);
``` 
This is incorrect, and may cause runtime errors and possibly segfault. Instead of the size of the linear system `n_`, the Gram-Schmidt class needs to be allocated with respect to the size of the sketching subspace:
```c++ 
    GS_->setup(k_rand_, restart_);
```

CC @kswirydo @shakedregev

### LINKED ISSUE #249: Refactor examples for `SystemSolver` class
### Summary

Use `ExampleHelper` class to declutter examples showing how to use `SystemSolver` class. Make examples portable, so they can run on multiple GPUs and we don't need to have separate examples for HIP and CUDA. Make sure examples file names follow Re::Solve style guidelines.

### Rationale

Currently `SystemSolver` examples compute different error norms in the body of the examples main functions. That takes up more code than the code that instantiates and runs the `SystemSolver` itself. Using `ExampleHelper` class takes utility functions outside of the example main body and makes examples more educational.

### Description

Branch off `refactor-examples-dev` and make a PR back to that branch.

Use [gpuRefactor](https://github.com/ORNL/ReSolve/blob/refactor-examples-dev/examples/gpuRefactor.cpp) as an example how to refactor `r_SysSolverCuda.cpp`, `r_SysSolverHip.cpp` and `r_SysSolverHipRefine.cpp`. All three examples should be produced from a single source file (suggested name `sysRefactor`).

### Additional information

_No response_

pelesh:
Create `ExampleHelper` class to encapsulate some common functions used in Re::Solve examples, so that examples are less cluttered and more informative.

Refactored examples are portable and each provides option of using iterative refinement, so we reproduce old examples with fewer source files. Experimental examples that are of interest only to Re::Solve developers are moved to a separate directory.

CC @kswirydo 

Some outstanding items before this draft is ready for review:
- [x] Expand functionality in `gpuRefactor.cpp` so that the example can use CUDA SDK and ROCm solvers.
- [x] Implement `sysRefactor.cpp` example using `ExampleHelper` class.
- [x] Implement GMRES example(s) using `ExampleHelper` class.
- [x] Implement `kluRefactor.cpp` for refactorization with iterative refinement on CPU.
- [x] Instrument examples for performance profiling.

Also: 
- Closes #221
- Closes #235
- Closes #249


shakedregev:
Addressed the first task [here](https://github.com/ORNL/ReSolve/pull/234)

shakedregev:
Unchecked the completed tasks as they are failing. See issues [255](https://github.com/ORNL/ReSolve/issues/255) and [256](https://github.com/ORNL/ReSolve/issues/256)

shakedregev (approved):
I reviewed every single one of the merges to this branch when they occurred and now tested the entire thing. All tests pass for CUDA and HIP

shakedregev:
When I rebase this with respect to `develop` to fix the merge conflict, the matrix handler test fails.

shakedregev (changes requested):
There is currently a merge conflict. When I fix it by rebasing, test 35 fails.

pelesh:
> There is currently a merge conflict. When I fix it by rebasing, test 35 fails on Frontier.

Branch rebased to `develop`. All tests pass and examples behave as expected.

shakedregev (approved):
Approved with minor comments

PR #228, opened 2025-02-13
Update GPU profiling interface

### LINKED ISSUE #195: Transition to rocprofiler-sdk
Per @jrmadsen, AMD's support for roctracer has ended. We should transition Re::Solve code to rocprofiler-sdk. See below.

CC: @PaulMullowney @kswirydo 

>Hi @ax3l @naromero77amd @pelesh @cameronrutherford @upsj just wanted to provide an update here. We have pretty much stopped development on roctracer, rocprofiler, and rocprofiler v2. We designed a new profiling library from scratch called [rocprofiler-sdk](https://github.com/rocm/rocprofiler-sdk) and did a beta release in ROCm 6.2. It is dramatically improved over the previous iterations. The API is relatively stable at this point and we’ve done hackathons with several tools at this point (TAU, Caliper, HPCToolkit, Linaro Map, CrayPAT, PAPI, etc.) and I believe y’all will be as happy with our new direction as they were. Y’all will be please to hear that
>```cmake
>find_package(rocprofiler-sdk)
>find_package(rocprofiler-sdk-roctx)
>```
>are fully supported with namespaced imported interface CMake targets. We are distributing a new ROCTx for various reasons: the old ROCTx couldn’t be separated from roctracer without breaking existing codes, you couldn’t install the old ROCTx without installing roctracer, etc. so this new ROCTx is designed to be standalone (in addition to adding new functions such as roctxProfilerPause and roctxProfilerResume).
>
>We encourage all of y’all to transition to using the new APIs sooner rather than later and if y’all need a consult on the transition, please let me know and we can set something up.

nkoukpaizan:
This updates the GPU profiling interfaces (roctx and nvtx):
- Makes use of the newer [rocprofiler-sdk](https://github.com/rocm/rocprofiler-sdk) for profiling on HIP platforms. This requires ROCm>6.2.0.
- Adds corresponding definitions for CUDA nvToolsExt.

Will need to update the documentation on profiling.
Fixes #195.


pelesh (changes requested):
Looks great, thanks @nkoukpaizan!

Could you please add instructions for using CUDA profiler to the Sphynx [documentation](https://github.com/ORNL/ReSolve/blob/nicholson/rocprofiler-sdk/docs/sphinx/developer_guide/profiling.rst)?

nkoukpaizan:
> Could you please add instructions for using CUDA profiler to the Sphynx [documentation](https://github.com/ORNL/ReSolve/blob/nicholson/rocprofiler-sdk/docs/sphinx/developer_guide/profiling.rst)?

Addressed in [99d380f](https://github.com/ORNL/ReSolve/pull/228/commits/99d380f6ac6b69805653e9e26cd3bad4fe82cfaf).

pelesh (approved):

PR #236, opened 2025-02-21
Fix bug where csc2csr method does not mark memory space updated

pelesh:
Method `csc2csr` sets CSR matrix data by accessing its raw data pointers. Therefore, when it changes the matrix data it should invoke method `setUpdated` to set the flag indicating that the matrix data is current. Otherwise, matrix will consider that data out of date and may overwrite it later. This PR fixes incorrectly set flags after `csc2csr` execution.

Also in this PR:
- Removed debugging outputs.
- Fixed code style.
- Updated formatting of matrix handler tests output


shakedregev (approved):
Looks good. Tests pass.

PR #239, opened 2025-02-24
Added transpose functionality and tests

shakedregev:
The tests reuse the formulas from the `csc2csr` test and pass Cpu, Cuda, and Hip.
All functions are documented, though tests are not (to be addressed separately).

### DIFF
--- a/resolve/matrix/MatrixHandlerCpu.hpp
+++ b/resolve/matrix/MatrixHandlerCpu.hpp
@@ -4,7 +4,7 @@
 #include <resolve/matrix/MatrixHandlerImpl.hpp>
 
 namespace ReSolve
-{ 
+{
   namespace vector
   {
     class Vector;
@@ -23,13 +23,13 @@ namespace ReSolve
 namespace ReSolve {
   /**
    * @class MatrixHandlerCpu
-   * 
+   *
    * @brief CPU implementation of the matrix handler.
    */
   class MatrixHandlerCpu : public MatrixHandlerImpl
   {
     using vector_type = vector::Vector;
-    
+
     public:
       MatrixHandlerCpu();
       MatrixHandlerCpu(LinAlgWorkspaceCpu* workspace);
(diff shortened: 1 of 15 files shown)


resolve/matrix/MatrixHandlerCpu.hpp
```diff
@@ -23,29 +23,31 @@ namespace ReSolve
...
                  vector_type* vec_x,
                  vector_type* vec_result,
                  const real_type* alpha,
                  const real_type* beta);
       virtual int matrixInfNorm(matrix::Sparse *A, real_type* norm);
-      void setValuesChanged(bool isValuesChanged); 
-    
-    private: 
+      void setValuesChanged(bool isValuesChanged);
+
```
pelesh, line 49: This still needs to be a public method.

resolve/matrix/MatrixHandlerHip.cpp
```diff
@@ -237,4 +237,58 @@ namespace ReSolve {
...
   }
 
+  /**
+   * @brief Transpose a sparse CSR matrix in HIP
+   *
+   * @param[in]  A - Sparse matrix
+   * @param[out] At - Transposed matrix
+   * @return int error_sum, 0 if successful
+   */
+  int MatrixHandlerHip::transpose(matrix::Csr* A, matrix::Csr* At)
```
pelesh, line 247:
```suggestion
   * @return int error_sum, 0 if successful
   *
   * @warning This method works only for `real_type == double`.
   */
  int MatrixHandlerHip::transpose(matrix::Csr* A, matrix::Csr* At)
```

pelesh (approved):
Looks good!

One suggestion: To make unit tests consistent, function `verifyCsrMatrix` in `MatrixHandlerTests` class should return `bool` and not take `TestStatus& status` parameter as an argument. This is a bit of an overkill. The function should return `false` and print appropriate message if the verification does not pass, but it does not have to provide detailed diagnostics.

PR #245, opened 2025-03-08
Fix bug in Gram-Schmidt test

pelesh:
In `GramSchmidtTests::orthogonalize`, the `GramSchmidt` class was set up as
```c++
          GS.setup(N, restart + 1);
```
instead of as
```c++
          GS.setup(N, restart);
```
That was causing the test to segfault. The workaround used before this PR was to allocate larger Hessenberg matrix. 

In this PR:
- `GramSchmidt` class is correctly set up.
- "Magic numbers" were replaced by variables with intuitive names to improve code readability.
- Additional comments were added to the code to explain what the test is doing.
- Test runner was refactored to remove unnecessary code duplication.

Verified:
- [x] Code builds without warnings with flags `-Wall -Wconversion -Wextra`
- [x] CPU tests pass
- [x] CUDA tests pass
- [ ] HIP tests pass

### DIFF
--- a/tests/unit/vector/GramSchmidtTests.hpp
+++ b/tests/unit/vector/GramSchmidtTests.hpp
@@ -113,9 +113,10 @@ namespace ReSolve
           nrm = 1.0 / nrm;
           handler_.scal(&nrm, &V, memspace_);
 
+          // Orthogonalize system and verify result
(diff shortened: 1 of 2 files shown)


shakedregev (changes requested):
Tests pass.

shakedregev (approved):

PR #246, opened 2025-03-10
Added feature: reusing transpose allocation

shakedregev:
Added the ability to reuse a transpose allocation by passing a corresponding `allocated_` flag.

This feature fits nicely with the working thesis of ReSolve, that memory should not be allocated unnecessarily, and reused between in subsequent linear iterations in the solution of the same nonlinear system.
This is crucial for HyKKT, which requires explicit access to the transpose of some matrices, and may be useful for other methods.

Added tests for CPU, CUDA, and HIP and all tests pass.

### DIFF
--- a/resolve/matrix/MatrixHandler.cpp
+++ b/resolve/matrix/MatrixHandler.cpp
@@ -214,38 +214,18 @@ namespace ReSolve {
    *
    * @return 0 if successful, 1 otherwise
    */
-  int MatrixHandler::transpose(matrix::Csr* A, matrix::Csr* At, memory::MemorySpace memspace, bool allocated)
+  int MatrixHandler::transpose(matrix::Csr* A, matrix::Csr* At, memory::MemorySpace memspace)
   {
     using namespace ReSolve::memory;
     switch (memspace) {
       case HOST:
-        return cpuImpl_->transpose(A, At, allocated);
+        return cpuImpl_->transpose(A, At);
         break;
       case DEVICE:
-        return devImpl_->transpose(A, At, allocated);
+        return devImpl_->transpose(A, At);
         break;
     }
     return 1;
-  }
-
-  /**
-   * @brief Add a constant to the nonzero values of a csr matrix.
-   * @param[in,out] A - Sparse matrix
-   * @param[in] alpha - scalar parameter
-   * @param[in] memspace - Device where the operation is computed
-   * @return 0 if successful, 1 otherwise
-   */
-  void MatrixHandler::addConstantToNonzeroValues(matrix::Sparse* A, real_type alpha, memory::MemorySpace memspace)
-  {
-    using namespace ReSolve::memory;
-    switch (memspace) {
-      case HOST:
-        cpuImpl_->addConstantToNonzeroValues(A, alpha);
-        break;
-      case DEVICE:
-        devImpl_->addConstantToNonzeroValues(A, alpha);
-        break;
-    }
   }
 
   /**
--- a/resolve/matrix/MatrixHandler.cpp
+++ b/resolve/matrix/MatrixHandler.cpp
@@ -214,15 +214,15 @@ namespace ReSolve {
    *
    * @return 0 if successful, 1 otherwise
    */
-  int MatrixHandler::transpose(matrix::Csr* A, matrix::Csr* At, memory::MemorySpace memspace)
+  int MatrixHandler::transpose(matrix::Csr* A, matrix::Csr* At, memory::MemorySpace memspace, bool allocated)
   {
     using namespace ReSolve::memory;
     switch (memspace) {
       case HOST:
-        return cpuImpl_->transpose(A, At);
+        return cpuImpl_->transpose(A, At, allocated);
         break;
       case DEVICE:
-        return devImpl_->transpose(A, At);
+        return devImpl_->transpose(A, At, allocated);
         break;
     }
     return 1;
--- a/resolve/matrix/MatrixHandler.cpp
+++ b/resolve/matrix/MatrixHandler.cpp
@@ -229,6 +229,26 @@ namespace ReSolve {
   }
 
   /**
+   * @brief Add a constant to the nonzero values of a csr matrix.
+   * @param[in,out] A - Sparse matrix
+   * @param[in] alpha - scalar parameter
+   * @param[in] memspace - Device where the operation is computed
+   * @return 0 if successful, 1 otherwise
+   */
+  void MatrixHandler::addConstantToNonzeroValues(matrix::Sparse* A, real_type alpha, memory::MemorySpace memspace)
+  {
+    using namespace ReSolve::memory;
+    switch (memspace) {
+      case HOST:
+        cpuImpl_->addConstantToNonzeroValues(A, alpha);
+        break;
+      case DEVICE:
+        devImpl_->addConstantToNonzeroValues(A, alpha);
+        break;
+    }
+  }
+
+  /**
    * @brief If CUDA support is enabled in the handler.
    *
    * @return true
--- a/resolve/matrix/MatrixHandler.cpp
+++ b/resolve/matrix/MatrixHandler.cpp
@@ -217,19 +217,27 @@ namespace ReSolve {
   int MatrixHandler::transpose(matrix::Csr* A, matrix::Csr* At, memory::MemorySpace memspace)
   {
     using namespace ReSolve::memory;
+    // check dimensions of A and At and if both are allocated
+    assert(A->getNumRows() == At->getNumColumns() && "Number of rows in A must be equal to number of columns in At");
+    assert(A->getNumColumns() == At->getNumRows() && "Number of columns in A must be equal to number of rows in At");
+    assert(A->getNnz() == At->getNnz() && "Number of nonzeros in A must be equal to number of nonzeros in At");
+    assert(A->getSparseFormat() == matrix::Sparse::COMPRESSED_SPARSE_ROW &&
+          "Matrix has to be in CSR format for transpose.\n");
+    assert(At->getSparseFormat() == matrix::Sparse::COMPRESSED_SPARSE_ROW &&
+          "Matrix has to be in CSR format for transpose.\n");
     switch (memspace) {
       case HOST:
+        if(A->getValues(memory::HOST) == nullptr) {
+          out::error() << "In MatrixHandler::transpose, A->getValues(memory::HOST) is null!\n";
+          return 1;
+        }
+        if(At->getValues(memory::HOST) == nullptr) {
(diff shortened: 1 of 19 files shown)


pelesh:
I converted this to draft as it needs more discussion.

shakedregev:
@pelesh - ready for review

resolve/matrix/MatrixHandlerCuda.cpp
```diff
@@ -317,19 +325,27 @@ namespace ReSolve {
...
+  /**
+   * @brief Add a constant to all nonzero values in the matrix
+   *
+   * @param[in, out] A - matrix
+   * @param[in] alpha - constant to be added
+   *
+   * @return int error code, 0 if successful
+   */
+  int MatrixHandlerCuda::addConstantToNonzeroValues(matrix::Sparse* A, real_type alpha)
+  {
```
pelesh, line 345: How about shortening this name?
```suggestion
  int MatrixHandlerCuda::addConstantToNonzeros(matrix::Sparse* A, real_type alpha)
  {
```

resolve/matrix/MatrixHandlerHip.cpp
```diff
@@ -238,59 +238,80 @@ namespace ReSolve {
   }
 
   /**
-   * @brief Transpose a sparse CSR matrix in HIP
+   * @brief Transpose a sparse CSR matrix.
+   *
```
pelesh, line 242:
```suggestion
   * @brief Transpose a sparse CSR matrix (HIP backend).
   *
```

resolve/matrix/MatrixHandlerHip.cpp
```diff
@@ -238,59 +238,80 @@ namespace ReSolve {
...
+    rocsparse_status status;
+    void* buffer_transpose = workspace_->getTransposeWorkspace();
+    bool allocated = workspace_->isTransposeAllocated();
+    if (!allocated) {
+      // check dimensions of A and At
+      assert(A->getNumRows() == At->getNumColumns() && "Number of rows in A must be equal to number of columns in At");
+      assert(A->getNumColumns() == At->getNumRows() && "Number of columns in A must be equal to number of rows in At");
 
-    status = rocsparse_csr2csc_buffer_size(workspace_->getRocsparseHandle(),
+      At->allocateMatrixData(memory::DEVICE);
```
pelesh, line 266: Whether matrix is allocated should be tracked in the matrix, not a workspace. We have to put more thought into how we want to handle this. Possible (quick and dirty) solutions are:
- Require that the input matrix `At` is always allocated with the same number of nonzeros as `A`. The user is responsible to maintain `At`. I am in favor of this.
- Check if matrix size and number of nonzeros matches `A`, and then make sure data pointers are of `At` are not null.
    - If this holds, reuse matrix.
    - If pointers are null, allocate matrix.
    - If sizes do not match exit function with error.

pelesh (changes requested):
We need to think a little bit more what preconditions on `At` are consistent with the rest of the Re::Solve code. Perhaps at the next hackathon?

resolve/matrix/MatrixHandlerCpu.cpp
```diff
@@ -233,10 +233,6 @@ namespace ReSolve
    */
   int MatrixHandlerCpu::transpose(matrix::Csr* A, matrix::Csr* At)
   {
-    assert(A->getNnz() == At->getNnz());
-    assert(A->getNumRows() == At->getNumColumns());
-    assert(A->getNumColumns() == At->getNumRows());
-
```
pelesh, line 239: We need to make sure that `At` is allocated and of the appropriate size.

resolve/matrix/MatrixHandlerCuda.cpp
```diff
@@ -266,43 +266,47 @@ namespace ReSolve {
...
+                                             At->getRowData(memory::DEVICE),
+                                             At->getColData(memory::DEVICE),
+                                             CUDA_R_64F,
+                                             CUSPARSE_ACTION_NUMERIC,
+                                             CUSPARSE_INDEX_BASE_ZERO,
+                                             CUSPARSE_CSR2CSC_ALG1,
+                                             &bufferSize);
+      error_sum += status;
+      mem_.allocateBufferOnDevice(&buffer_transpose, bufferSize);
+      workspace_->setTransposeAllocated();
```
pelesh, line 308: I don't think this is doing what it is intended to do. It is setting value of the local pointer to the array just allocated on the device. But, when the local pointer gets out of scope, there will be memory leak. The buffer pointer in the workspace will remain null, so I think each time you call this function you would allocate and leak memory.

resolve/matrix/MatrixHandlerHip.cpp
```diff
@@ -238,59 +238,78 @@ namespace ReSolve {
...
+                                           A->getColData(memory::DEVICE),
                                            rocsparse_action_numeric,
                                            &bufferSize);
-
-    error_sum += status;
-    mem_.allocateBufferOnDevice(&d_work, bufferSize);
-    
+      error_sum += status;
+      mem_.allocateBufferOnDevice(&buffer_transpose, bufferSize);
+      workspace_->setTransposeAllocated();
```
pelesh, line 277: Same as for the CUDA implementation -- allocated is local pointer, so when it goes out of scope, there will be a memory leak. The pointer in the workspace will remain null.

pelesh (changes requested):
The memory leak needs to be fixed in HIP and CUDA implementations of the `transpose` function.

resolve/workspace/LinAlgWorkspaceHIP.hpp
```diff
@@ -57,11 +60,13 @@ namespace ReSolve
 
       //info - but we need info
       rocsparse_mat_info info_A_;
-      
+
       real_type* d_r_{nullptr}; // needed for inf-norm
       real_type* norm_buffer_{nullptr}; // needed for inf-norm
+      real_type* transpose_workspace_{nullptr}; // needed for transpose
+      bool transpose_workspace_ready_{false}; // to track if allocated
```
pelesh, line 67: This is a bug. The workspace pointer must be of type `void*`.
```suggestion
      void* transpose_workspace_{nullptr}; // needed for transpose
      bool transpose_workspace_ready_{false}; // to track if allocated
```

shakedregev:
All tests pass. Ready for re-review.

resolve/matrix/MatrixHandler.cpp
```diff
@@ -222,12 +222,48 @@ namespace ReSolve {
...
         break;
       case DEVICE:
+            // check dimensions of A and At and if both are allocated
+        assert(A->getNumRows() == At->getNumColumns() && "Number of rows in A must be equal to number of columns in At");
+        assert(A->getNumColumns() == At->getNumRows() && "Number of columns in A must be equal to number of rows in At");
+        assert(A->getNnz() == At->getNnz() && "Number of nonzeros in A must be equal to number of nonzeros in At");
+        assert(A->getSparseFormat() == matrix::Sparse::COMPRESSED_SPARSE_ROW &&
+              "Matrix has to be in CSR format for transpose.\n");
+        assert(At->getSparseFormat() == matrix::Sparse::COMPRESSED_SPARSE_ROW &&
+              "Matrix has to be in CSR format for transpose.\n");
```
pelesh, line 232: None of these tests is specific to the device implementation. I suggest we have them _before_ the switch statement.

resolve/matrix/MatrixHandlerCpu.cpp
```diff
@@ -246,6 +244,10 @@ namespace ReSolve
     index_type* rowPtrAt = At->getRowData(memory::HOST);
     index_type* colIdxAt = At->getColData(memory::HOST);
     real_type*  valuesAt = At->getValues( memory::HOST);
+    // Check dimensions of A and At
+    assert(A->getNumRows() == At->getNumColumns() && "Number of rows in A must be equal to number of columns in At");
+    assert(A->getNumColumns() == At->getNumRows() && "Number of columns in A must be equal to number of rows in At");
+    assert(A->getNnz() == At->getNnz() && "Number of nonzeros in A must be equal to number of nonzeros in At");
```
pelesh, line 250: These checks could be moved to `MatrixHandler` class as they are not implementation specific. That would reduce overall size of the code in handlers.

resolve/matrix/MatrixHandlerCuda.cpp
```diff
@@ -266,43 +266,48 @@ namespace ReSolve {
...
-                                                            At->getColData(memory::DEVICE),
-                                                            CUDA_R_64F,
-                                                            CUSPARSE_ACTION_NUMERIC,
-                                                            CUSPARSE_INDEX_BASE_ZERO,
-                                                            CUSPARSE_CSR2CSC_ALG1,
-                                                            &bufferSize);
-    error_sum += status;
-    mem_.allocateBufferOnDevice(&d_work, bufferSize);
-
+    assert(A->getNnz() == At->getNnz() && "Number of nonzeros in A must be equal to number of nonzeros in At");
```
pelesh, line 289: Same here: These asserts could  (and probably should) be in `MatrixHandler` class.

resolve/matrix/MatrixHandlerHip.hpp
```diff
@@ -23,34 +23,35 @@ namespace ReSolve
...
       MatrixHandlerHip(LinAlgWorkspaceHIP* workspace);
       virtual ~MatrixHandlerHip();
 
       int csc2csr(matrix::Csc* A_csc, matrix::Csr* A_csr);
 
-      int transpose(matrix::Csr* A, matrix::Csr* At);
-      
+      int transpose(matrix::Csr* A, matrix::Csr* At) override;
+
+      int addConst(matrix::Sparse* A, real_type alpha);
```
pelesh, line 42: Why does `transpose` have `override` qualifier and `addConst` does not?

resolve/matrix/MatrixHandlerHip.cpp
```diff
@@ -238,59 +238,80 @@ namespace ReSolve {
...
-    void* d_work;
-
-    status = rocsparse_csr2csc_buffer_size(workspace_->getRocsparseHandle(),
+    rocsparse_status status;
+    void* buffer_transpose = workspace_->getTransposeBufferWorkspace();
+    bool allocated = workspace_->isTransposeBufferAllocated();
+    // check dimensions of A and At
+    assert(A->getNumRows() == At->getNumColumns() && "Number of rows in A must be equal to number of columns in At");
+    assert(A->getNumColumns() == At->getNumRows() && "Number of columns in A must be equal to number of rows in At");
+    assert(A->getNnz() == At->getNnz() && "Number of nonzeros in A must be equal to number of nonzeros in At");
```
pelesh, line 266: This is not implementation specific, so it fits better in `MatrixHandler` class.

PR #248, opened 2025-03-20
Correcting PR template

shakedregev:
## Description
 
 _Github does not support different PR templates. If more than one template is in the folder, it defaults to the blank template. Github will add support for multiple templates in the future, so we want to keep them, but make sure it recognizes our ReSolve specific (but otherwise generic) template. I also updated this basic template._
 

 ## Proposed changes
 
 _Moved specific PR templates for future use. Updated generic template. This allows Github to recognize the template._


pelesh (changes requested):
This PR avoids GitHub confusion with multiple PR templates. However, the currently active [template](https://github.com/ORNL/ReSolve/blob/shaked/single_pr_template/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md) is not very helpful. Instead, I would just gently modify bug report template to be a tad more general and use that as the default PR template. I think the checklist we have there would be quite helpful going forward.

shakedregev:
> This PR avoids GitHub confusion with multiple PR templates. However, the currently active [template](https://github.com/ORNL/ReSolve/blob/shaked/single_pr_template/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md) is not very helpful. Instead, I would just gently modify bug report template to be a tad more general and use that as the default PR template. I think the checklist we have there would be quite helpful going forward.

@pelesh - sorry about this. I made the change but somehow it didn't save before the push. Should be good now.

Issue #250, opened 2025-04-03
Refactor KLU examples


pelesh:
### Summary

Use ExampleHelper class to declutter examples showing how to use KLU from Re::Solve. Make sure examples file names follow Re::Solve style guidelines.

### Rationale

Currently KLU examples compute different error norms in the body of the examples main functions. That takes up more code than the code that instantiates and runs the solver itself. Using ExampleHelper class takes utility functions outside of the example main body and makes examples more educational.

### Description

Branch off refactor-examples-dev and make a PR back to that branch.

Use [gpuRefactor](https://github.com/ORNL/ReSolve/blob/refactor-examples-dev/examples/gpuRefactor.cpp) as an example how to refactor r_KLU_KLU.cpp, and r__KLU_KLU_standalone.cpp (suggested names `kluRefactor` and `kluStandalone`).

shakedregev:
Is there a good reason why there is a `r_SysSolverHipRefine.cpp` but not a `r_SysSolverCudaRefine.cpp`?

pelesh:
> Is there a good reason why there is a `r_SysSolverHipRefine.cpp` but not a `r_SysSolverCudaRefine.cpp`?

The right question is why are there separate `r_SysSolverHip.cpp` and `r_SysSolverCuda.cpp` 😄. Similar as in [functionality tests](https://github.com/ORNL/ReSolve/blob/develop/tests/functionality/testSysRefactor.cpp), these examples should be produced by the same source file also with an option to turn the iterative refinement on and off as needed.

PR #253, opened 2025-04-07
Refactor and stand alone examples for KLU class

shakedregev:
This PR is an enhancement solving [Issue 250](https://github.com/ORNL/ReSolve/issues/250)

`kluRefactor.exe` and `kluStandAlone.exe` are now decluttered using `ExampleHelper` and `CliOptions`. I kept the two original files, but they are not needed.

Tests pass on CPU.

### DIFF
--- a/examples/kluRefactor.cpp
+++ b/examples/kluRefactor.cpp
@@ -72,7 +72,7 @@ int kluRefactor(int argc, char *argv[])
   workspace.initializeHandles();
   // Create a helper object (computing errors, printing summaries, etc.)
   ReSolve::examples::ExampleHelper<workspace_type> helper(workspace);
-  std::cout << "sysRefactor with " << helper.getHardwareBackend() << " backend\n";
+  std::cout << "kluRefactor with " << helper.getHardwareBackend() << " backend\n";
 
   ReSolve::MatrixHandler matrix_handler(&workspace);
   ReSolve::VectorHandler vector_handler(&workspace);
--- a/examples/kluRefactor.cpp
+++ b/examples/kluRefactor.cpp
@@ -13,57 +13,68 @@
 #include <resolve/LinSolverDirectKLU.hpp>
 #include <resolve/workspace/LinAlgWorkspace.hpp>
 #include "ExampleHelper.hpp"
+#include <resolve/utilities/params/CliOptions.hpp>
 
 using namespace ReSolve::constants;
 
-/// Prototype of the example function
-template <class workspace_type, class refactor_type>
-static int kluRefactor(int argc, char *argv[]);
-
-/// Main function selects example to be run.
-int main(int argc, char *argv[])
+/// Prints help message describing system usage.
+void printHelpInfo()
 {
-  kluRefactor<ReSolve::LinAlgWorkspaceCpu,
-             ReSolve::LinSolverDirectKLU>(argc, argv);
-  #ifdef RESOLVE_USE_CUDA
-    kluRefactor<ReSolve::LinAlgWorkspaceCUDA,
-               ReSolve::LinSolverDirectKLU>(argc, argv);
-  #endif
-
-  #ifdef RESOLVE_USE_HIP
-    kluRefactor<ReSolve::LinAlgWorkspaceHIP,
-                ReSolve::LinSolverDirectKLU>(argc, argv);
-  #endif
-
-  return 0;
+  std::cout << "\nLoads from files and solves a series of linear systems.\n\n";
+  std::cout << "System matrices are in files with names <pathname>XX.mtx, where XX are\n";
+  std::cout << "consecutive integer numbers 00, 01, 02, ...\n\n";
+  std::cout << "System right hand side vectors are stored in files with matching numbering.\n";
+  std::cout << "and file extension.\n\n";
+  std::cout << "Usage:\n\t./";
+  std::cout << "kluRefactor.exe -m <matrix pathname> -r <rhs pathname> -n <number of systems>\n\n";
+  std::cout << "Optional features:\n\t-h\tPrints this message.\n";
+  std::cout << "\t-i\tEnables iterative refinement.\n\n";
 }
 
-/**
- * @brief Example of using refactorization solvers on for KLU
- *
- * @tparam workspace_type - Type of the workspace to use
- * @param[in] argc - Number of command line arguments
- * @param[in] argv - Command line arguments
- * @return 0 if the example ran successfully, -1 otherwise
- */
-template <class workspace_type, class refactor_type>
-int kluRefactor(int argc, char *argv[])
+int main(int argc, char *argv[])
 {
   // Use the same data types as those you specified in ReSolve build.
   using index_type = ReSolve::index_type;
   using real_type  = ReSolve::real_type;
   using vector_type = ReSolve::vector::Vector;
 
-  if (argc < 4) {
-    std::cerr << "Error: Not enough input parameters. Expected 3 parameters: <matrixFileName> <rhsFileName> <numSystems>" << std::endl;
-    return -1;
-  }
-  std::string  matrixFileName = argv[1];
-  std::string  rhsFileName = argv[2];
+  ReSolve::CliOptions options(argc, argv);
+  ReSolve::CliOptions::Option* opt = nullptr;
 
-  index_type numSystems = std::stoi(argv[3]);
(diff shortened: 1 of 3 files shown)


examples/kluRefactor.cpp
```diff
@@ -0,0 +1,196 @@
...
+  #ifdef RESOLVE_USE_CUDA
+    kluRefactor<ReSolve::LinAlgWorkspaceCUDA,
+               ReSolve::LinSolverDirectKLU>(argc, argv);
+  #endif
+
+  #ifdef RESOLVE_USE_HIP
+    kluRefactor<ReSolve::LinAlgWorkspaceHIP,
+                ReSolve::LinSolverDirectKLU>(argc, argv);
+  #endif
+
```
pelesh, line 37: We don't need to do this for KLU example, it is all on CPU.

examples/kluRefactor.cpp
```diff
@@ -0,0 +1,196 @@
...
+  // Use the same data types as those you specified in ReSolve build.
+  using index_type = ReSolve::index_type;
+  using real_type  = ReSolve::real_type;
+  using vector_type = ReSolve::vector::Vector;
+
+  if (argc < 4) {
+    std::cerr << "Error: Not enough input parameters. Expected 3 parameters: <matrixFileName> <rhsFileName> <numSystems>" << std::endl;
+    return -1;
+  }
+  std::string  matrixFileName = argv[1];
```
pelesh, line 61: Use `CliOptions` class to parse command line input. Also, a good idea would be to add usage instructions.

examples/kluRefactor.cpp
```diff
@@ -0,0 +1,196 @@
...
+  std::string rhsId;
+  std::string matrixFileNameFull;
+  std::string rhsFileNameFull;
+
+  ReSolve::matrix::Csr* A = nullptr;
+  workspace_type workspace;
+  workspace.initializeHandles();
+  // Create a helper object (computing errors, printing summaries, etc.)
+  ReSolve::examples::ExampleHelper<workspace_type> helper(workspace);
+  std::cout << "kluRefactor with " << helper.getHardwareBackend() << " backend\n";
```
pelesh, line 78: Again, KLU is a CPU-based solver, hardware backend is always CPU.

examples/kluRefactor.cpp
```diff
@@ -0,0 +1,196 @@
...
+    rhsFileNameFull = "";
+
+    // Read matrix first
+    matrixFileNameFull = matrixFileName + fileId + ".mtx";
+    rhsFileNameFull = rhsFileName + rhsId + ".mtx";
+    std::cout << std::endl << std::endl << std::endl;
+    std::cout << "========================================================================================================================"<<std::endl;
+    std::cout << "Reading: " << matrixFileNameFull << std::endl;
+    std::cout << "========================================================================================================================"<<std::endl;
+    std::cout << std::endl;
```
pelesh, line 108: The whole purpose of using `ExampleHelper` is to declutter code and remove implementation of output formatting from the example main body.

examples/kluRefactor.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+    std::cout << "\t2-Norm of the residual: "
+              << std::scientific << std::setprecision(16)
+              << sqrt(vector_handler.dot(vec_r, vec_r, ReSolve::memory::HOST)) << "\n";
+    matrix_handler.matrixInfNorm(A, &norm_A, ReSolve::memory::HOST);
+    norm_x = vector_handler.infNorm(vec_x, ReSolve::memory::HOST);
+    std::cout << "\tMatrix inf  norm: " << std::scientific << std::setprecision(16) << norm_A<<"\n"
+              << "\tResidual inf norm: " << norm_r <<"\n"
+              << "\tSolution inf norm: " << norm_x <<"\n"
+              << "\tNorm of scaled residuals: "<< norm_r / (norm_A * norm_x) << "\n";
```
pelesh, line 183: Again, the idea of using ExampleHelper is to remove code like this from the main body of the example and declutter the code.

examples/kluRefactor.cpp
pelesh: See e.g. [`gpuRefactor.cpp`](https://github.com/ORNL/ReSolve/blob/refactor-examples-dev/examples/gpuRefactor.cpp).

pelesh (changes requested):
The KLU examples are not really decluttered and they do not use `ExampleHelper` class. Also, `CliOptions` class should be used to process command line input.

examples/kluRefactor.cpp
```diff
@@ -0,0 +1,180 @@
...
+      status =  KLU->refactorize();
+      std::cout<<"KLU re-factorization status: "<<status<<std::endl;
+    }
+    status = KLU->solve(vec_rhs, vec_x);
+    std::cout<<"KLU solve status: "<<status<<std::endl;
+    vec_r->copyDataFrom(rhs, ReSolve::memory::HOST, ReSolve::memory::HOST);
+
+    matrix_handler->setValuesChanged(true, ReSolve::memory::HOST);
+
+    matrix_handler->matvec(A, vec_x, vec_r, &ONE, &MINUSONE, ReSolve::memory::HOST);
```
pelesh, line 163: What is the purpose of these functions?

examples/kluRefactor.cpp
```diff
@@ -0,0 +1,180 @@
...
+      vec_rhs->copyDataFrom(rhs, ReSolve::memory::HOST, ReSolve::memory::HOST);
+    }
+    std::cout<<"COO to CSR completed. Expanded NNZ: "<< A->getNnz()<<std::endl;
+    //Now call direct solver
+    int status;
+
+    if (i < 2){
+      KLU->setup(A);
+      status = KLU->analyze();
+      std::cout<<"KLU analysis status: "<<status<<std::endl;
```
pelesh, line 150: Why is condition here `i < 2`? Shouldn't one factorization be sufficient to start refactorization series?

examples/kluRefactor.cpp
```diff
@@ -0,0 +1,180 @@
...
+      status = KLU->factorize();
+      std::cout<<"KLU factorization status: "<<status<<std::endl;
+    } else {
+      status =  KLU->refactorize();
+      std::cout<<"KLU re-factorization status: "<<status<<std::endl;
+    }
+    status = KLU->solve(vec_rhs, vec_x);
+    std::cout<<"KLU solve status: "<<status<<std::endl;
```
pelesh, line 158: It would be good to provide an option for iterative refinement in this example.

examples/kluRefactor.cpp
```diff
@@ -0,0 +1,180 @@
...
+    {
+      std::cout << "Failed to open file " << rhs_file_name_full << "\n";
+      return -1;
+    }
+    bool is_expand_symmetric = true;
+    if (i == 0) {
+      A = ReSolve::io::createCsrFromFile(mat_file, is_expand_symmetric);
+
+      rhs = ReSolve::io::createArrayFromFile(rhs_file);
+      x = new real_type[A->getNumRows()];
```
pelesh, line 124: We can create vector directly from file. That way we reduce clutter in this example.

examples/kluRefactor.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+  bool is_iterative_refinement = options.hasKey("-i");
+
+  index_type num_systems = 0;
+  opt = options.getParamFromKey("-n");
+  if (opt) {
+    num_systems = std::stoi((opt->second).c_str());
+  } else {
+    std::cout << "Incorrect input!\n";
+    printHelpInfo();
```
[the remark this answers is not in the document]
pelesh: Good suggestion but perhaps better to simply return rather than `exit(1)`.

pelesh (changes requested):
Nice progress. A few changes requested before we can merge this.

examples/kluRefactor.cpp
shakedregev: In theory yes, in practice it doesn't work. You get `nan`s. It could be something pathological about the test set, e.g. the 0-th matrix is easier and doesn't require as much pivoting as the rest of the series.

shakedregev:
> Nice progress. A few changes requested before we can merge this.

@pelesh - made the necessary changes. This is ready

pelesh (approved):
Looks good, thank you!

PR #257, opened 2025-04-18
Improve error messages in sparse matrix classes

### LINKED ISSUE #229: Change errors Csc, Csr, Coo classes to give suggestions where possible
For example, change:
 ```c++
        if (!d_data_updated_) {
          out::error() << "In Csr::syncData trying to sync host with device, but device is out of date!\n";
          assert(d_data_updated_);
        }
```
 ```c++
        if (!d_data_updated_) {
          out::error() << "In Csr::syncData trying to sync host with device, but device is out of date! 
          If you have changed the data on purpose, update the device with "variableName->setUpdated(memory::DEVICE)". 
          If you did not mean to change the data on the device, check your code. \n";
          assert(d_data_updated_);
        }
```
@pelesh

pelesh:
Several changes are made to improve error messages returned to user:
- If one of arrays with row or column indices is allocated but the other is not, this is probably a bug in Re::Solve. Instead of sending error message to the user, use `assert` statement to catch such errors in CI pipeline.
- Send warning to the user if trying to sync memory space that is already up-to-date. In that case, the user is probably doing something wrong but the issue might be harmless. The code ignores command and continues to run.
- Send an error message to the user if they are trying to sync a memory space with the other that is out of date. This is most likely a bug in user's code; user is directed to `syncData` documentation.

Closes #229

### DIFF
--- a/resolve/matrix/Sparse.cpp
+++ b/resolve/matrix/Sparse.cpp
@@ -203,7 +203,7 @@ namespace ReSolve {
    * In such case, the matrix has no way of knowing which data is most recent, so
    * you have to tell it.
    * 
-   * @warning This is expert-level function. Use only if you know what you are
+   * @warning This is an expert-level function. Use only if you know what you are
    * doing.
    * 
    * @note If you want to set both DEVICE and HOST memory to the same value
--- a/resolve/matrix/Sparse.cpp
+++ b/resolve/matrix/Sparse.cpp
@@ -191,7 +191,7 @@ namespace ReSolve {
   }
 
   /**
-   * @brief Tags `memsapce` as updated.
+   * @brief Tags `memspace` as updated.
    *
    * @param[in] memspace - memory space (HOST or DEVICE) to set to "updated"
    *
--- a/resolve/matrix/Sparse.cpp
+++ b/resolve/matrix/Sparse.cpp
@@ -197,10 +197,10 @@ namespace ReSolve {
    *
    * @return 0 if successful, -1 if not.
    * 
-   * The method sets boolean flag indicating that the `memspace` is updated.
+   * The method sets the boolean flag indicating that the `memspace` is updated.
    * It automatically sets the other data mirror to non-updated. You would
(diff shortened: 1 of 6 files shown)


resolve/matrix/Sparse.cpp
```diff
@@ -191,13 +191,23 @@ namespace ReSolve {
   }
 
   /**
-   * @brief Set the data to be updated on HOST or DEVICE. 
+   * @brief Tags `memsapce` as updated.
```
shakedregev, line 194:
```suggestion
   * @brief Tags `memspace` as updated.
```

resolve/matrix/Sparse.cpp
```diff
@@ -191,13 +191,23 @@ namespace ReSolve {
...
+   * @param[in] memspace - memory space (HOST or DEVICE) to set to "updated"
    *
    * @return 0 if successful, -1 if not.
    * 
-   * @note The method automatically sets the other mirror data to non-updated (but it does not copy).
+   * The method sets boolean flag indicating that the `memspace` is updated.
+   * It automatically sets the other data mirror to non-updated. You would
+   * use this function if you update matrix data by accessing its raw pointers.
+   * In such case, matrix has no way of knowing which data is most recent, so
+   * you have to tell it.
```
shakedregev, line 204:
```suggestion
   * The method sets the boolean flag indicating that the `memspace` is updated.
   * It automatically sets the other data mirror to non-updated. You would
   * use this function if you update matrix data by accessing its raw pointers.
   * In such case, the matrix has no way of knowing which data is most recent, so
   * you have to tell it.
```

resolve/matrix/Sparse.cpp
```diff
@@ -191,13 +191,23 @@ namespace ReSolve {
...
+   * 
+   * @warning This is expert-level function. Use only if you know what you are
+   * doing.
```
shakedregev, line 207:
```suggestion
   * @warning This is an expert-level function. Use only if you know what you are
   * doing.
```

PR #258, opened 2025-04-19
Update pull request and issue templates.

pelesh:
Minor updates to templates for GitHub issues and pull requests.


shakedregev (approved):
Approved but please do not delete `slaven/update_issue_templates_dev` because we need copies of the specific PR templates for once Github enables that.

PR #260, opened 2025-04-19
Track updated status for each vector in multivector.

### LINKED ISSUE #216: Vector initialization functions do not set "updated" flags correctly

CC @kswirydo

### LINKED ISSUE #259: `Vector::getData` function incorrectly syncs memory spaces.
## Description of the issue

```c++
  {
      syncData(memspace);  
      owns_cpu_data_ = true;
    } 

  // other code ...
```

## Suggested solution

pelesh:

This PR provides following:
- Separate bookkeeping of host and device memory updates for each vector in the multivector object. Closes #259.


resolve/vector/Vector.cpp
```diff
@@ -280,13 +307,32 @@ namespace ReSolve { namespace vector {
   {
     using namespace ReSolve::memory;
 
+    bool all_gpu_updated = gpu_updated_[0];
```
kswirydo, line 310: I'm not sure if I follow the logic here. If the length of `gpu_updated` and 'cpu_updated' is k, shouldnt we just check every vector separately and copy, if needed?

kswirydo (commented):
First -- this is not a bug. What you are adding is a new `feature`. There is nothing wrong with current approach except ... it is limiting in some cases.

shakedregev (approved):
Looks good and tests pass, but maybe we want a test for this specifically? Separate PR?

resolve/vector/Vector.cpp
pelesh: Good question! This function syncs the entire multivector at once. It can do that only if all vectors in the multivector have the same updated status. Otherwise, it should exit with an error. Next few lines check if it is the case. After that, it is safe to use `all_gpu_updated`.

The check itself may add some overhead. Perhaps it should be enabled only in debug mode?

kswirydo: Okay, I see what you are doing.

PR #265, opened 2025-05-08
Enhancement: remove unnecessary second analyze 

shakedregev:
## Description
 
`kluRefactor`, `sysRefactor`, and `gpuRefactor` were all performing a redundant second analyze.
 
 ## Proposed changes
 
 _Made sure analyze is only done once. All tests pass._
 
 ## Checklist
 
 merging your code._
 
- [X] All tests pass. Code tested on
     - [X] CPU backend
     - [X] CUDA backend
     - [X] HIP backend
- [X] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [X] The new code follows Re::Solve style guidelines.
- [X] There are unit tests for the new code.
- [X] The new code is documented.

### DIFF
--- a/examples/gpuRefactor.cpp
+++ b/examples/gpuRefactor.cpp
@@ -228,8 +228,6 @@ int gpuRefactor(int argc, char *argv[])
         index_type* Q = KLU.getQOrdering();
 
         Rf.setup(A, L, U, P, Q, vec_rhs);
-        // Refactorization
-        // Rf.refactorize();
 
         // Setup iterative refinement solver
         if (is_iterative_refinement) {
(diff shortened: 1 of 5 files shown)


pelesh:
So with this modification, the solution to system `01` fails for rocsolverRf and for cusolverRf. As expected, cusolverGLU works fine here. This is probably a bug in `SystemSolver` class when it decides to run triangular solve on GPU.

pelesh (approved):
The fix to `SystemSolver` class was trivial so I pushed it here directly. Everything works fine now.

Issue #274, opened 2025-05-18
Bug: What is correct use of `n_current_` with multivectors?


pelesh:
### Description how to reproduce the bug

Public method `setCurrentSize(index_type n_current)` modifies `Vector` class member variable `n_current_` without reallocating the vector. This operation will be performed only if `n_current_ <= n_`, where `n_` is the size of the vector data array.

It seems that some of the other methods in `Vector` class are not using `n_current_` correctly. For example, in method `copyDataFrom` there is a line like this:
```c++
  mem_.copyArrayHostToHost(h_data_, data, n_current_ * k_);
```
Since for the multivector memory block is allocated to size `n_*k_`, vectors in the multivector start at `h_data_`, `h_data_ + n_`, `h_data_ + 2*n_`, ... . Copying `n_current_*k_` elements from `data` to the vector memory will not distribute (sub)arrays of `data` to matching vectors in the multivector, if `n_current_ < n_`.

Similar issues exist in `Vector` methods:
- `getData`
- `syncData`
- `setToZero`
- `setToConst`

It seems the only reason why this issue is not affecting computations in Re::Solve is because `setCurrentSize` is used in only two places in the code -- modified Gramm-Schmidt post-modern and 2-sync orthogonalizations. 

### Re::Solve version

`develop` branch at e42ba168aee4d5ad09b5330ffbebef2d21839d16

### System and environment details

AMD EPYC 7763 64-Core Processor, MI250X GPU, AMD clang version 18.0.0 on Linux 6.4.0.

pelesh:
Probably best solution here would be to remove `n_current_` member from the `Vector` class altogether and implement the two modified Gram-Schmidt variants without it.

shakedregev:
Why are current tests not catching the issue? Do none of the tests use Gram-Schmidt?

pelesh:
Another question: Should `Vector` class allow for resizing (mulit)vectors who do not own their data? If we allow that, then the owner of the data might be left in an ill-defined state after a vector resize and subsequent operations on the resized vector.

PR #275, opened 2025-05-20
Simplify LinSolverIterativeRandFGMRES

pelesh:
## Description

Class  `LinSolverIterativeRandFGMRES` uses local variable `vec_z` for two different purposes -- as a vector view for a select vector in multivector `vec_Z_` and as a vector view for an auxiliary data array `d_aux_`. The two different uses of `vec_z` make the code difficult to read. In such implementation, a couple of calls to `setData` and `setCurrentSize` are required. These methods (as implemented now) are error prone and should be avoided, if possible.

This was also mentioned in #274.

 ## Proposed changes
 
Instead of allocating member data array `d_aux_` and then attaching a vector view `vec_z` to it, one can simply create and allocate a member vector `vec_aux_` and use it instead. This simplifies the code as it eliminates unnecessary `setData` call, removes one `setCurrentSize`call, and does not require any additional memory allocation.

 
 ## Checklist
 
- [X] All tests pass. Code tested on
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 
 ## Further comments
 
When copying data from Hessenberg matrix to the auxiliary data array `d_aux_`, there are `i+2` elements copied while the vector view of the data array is configured with `i+1` elements. It seems only `i+1` elements are used later on in the computation. This PR does not address this discrepancy.

### DIFF
--- a/resolve/LinSolverIterativeRandFGMRES.cpp
+++ b/resolve/LinSolverIterativeRandFGMRES.cpp
@@ -241,18 +241,11 @@ namespace ReSolve
         mem_.deviceSynchronize();
         GS_->orthogonalize(k_rand_, vec_S_, h_H_, i);
         // now post-process
-        if (memspace_ == memory::DEVICE) {
-          mem_.copyArrayHostToDevice(d_aux_, &h_H_[i * (restart_ + 1)], i + 2);
-        } else {
-          mem_.copyArrayHostToHost(d_aux_, &h_H_[i * (restart_ + 1)], i + 2);
-        }
-        vec_z->setData(d_aux_, memspace_);
-        vec_z->setCurrentSize(i + 1);
+        vec_aux_->copyDataFrom(&h_H_[i * (restart_ + 1)], memory::HOST, memspace_);
+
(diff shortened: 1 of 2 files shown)


shakedregev:
Is there a compelling reason why we should have both `vec_v` and `vec_z`, but also `vec_V` and `vec_Z`? If these are different variables, they should have distinguishable names.

pelesh:
> Is there a compelling reason why we should have both `vec_v` and `vec_z`, but also `vec_V` and `vec_Z`? If these are different variables, they should have distinguishable names.

My understanding is that `vec_v` and `vec_z` are vector "views", created on stack as local variables, for the individual vectors in member multivectors `vec_V_` and `vec_Z_`.

shakedregev:
Right, I forgot the trailing underscore. In that case, ReSolve guidelines say that `vec_V` and `vec_Z` should be uncapitalized. These are not matrices, so Householder would also have them as lower case. If for some reason we want them uppercase, their views should be as well. I will create a PR to this PR as there are quite a few names here that do not follow conventions.

shakedregev (approved):
The new code looks good and tests pass on CUDA and HIP. 
Approving and recommending an eventual wholesale adapting of the names here to fit ReSolve guidelines.

PR #277, opened 2025-05-21
Adjust resizing in vector::Vector class

### LINKED ISSUE #274: Bug: What is correct use of `n_current_` with multivectors?

pelesh:
## Description
 
This PR addresses several minor issues in `vector::Vector` class:
- Changes ambiguous member variable and function names.
- `getSize()` now returns vector length rather than the size of allocated memory (vector capacity)
- Default values to data ownership flags are set to `true` now.
- Resizing of vectors who do not own data (i.e. vector views of external data) is not allowed anymore.
- `getData` does not set the data ownership.

Closes #274
 

 ## Proposed changes
 
### Renamed:

Old        |    New
--------|---------
`setCurrentSize`       |  `resize`
`n_`                            | `n_capacity_`
`n_current_`             |  `n_size_`
`getSize`                   | `getCapacity`
`getCurrentSize`    | `getSize`

### Changed defaults:

Member boolean variables `owns_gpu_data_` and `owns_cpu_data_` have default values `true`. They are set to `false` only when `setData` method is invoked. Therefore `false` values of these flags indicate that the vector object is used as a "view" of data external to the vector.

Resizing of vectors who do not own their data is not allowed anymore. The owner of the data should be the only one responsible for resizing it.

### Data ownership

Public method `getData(MemorySpace m)` has the ability to bring the data to the memory space requested (host or device). It now relies on `syncData` method to copy data to the requested memory space and set correct data ownership flags, if needed.

 
 ## Checklist
  
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 
 ## Further comments

### DIFF
--- a/resolve/vector/Vector.cpp
+++ b/resolve/vector/Vector.cpp
@@ -47,7 +47,7 @@ namespace ReSolve { namespace vector {
    * @brief Get capacity of a single vector.
    * 
    * Vector memory is allocated to `n_capacity_*k_`. This is the maximum
-   * size of elements that (multi)vector can hold.
+   * number of elements that the (multi)vector can hold.
    * 
    * @return `n_capacity_` maximum number of elements in the vector.
    */
--- a/resolve/vector/Vector.cpp
+++ b/resolve/vector/Vector.cpp
@@ -482,6 +482,8 @@ namespace ReSolve { namespace vector {
    * Use for vectors and multivectors that change size throughout computation.
    * 
    * @note Vector needs to have capacity set to maximum expected size.
+   * @warning This method is not to be used in vectors who do not own their
+   * data.
    * 
    * @param[in] new_n_size - New vector length
    *
@@ -493,9 +495,7 @@ namespace ReSolve { namespace vector {
   {
     assert(owns_cpu_data_ && owns_gpu_data_ 
            && "Cannot resize if vector is not owning the data.");
-    // if (!owns_cpu_data_ || !owns_gpu_data_) {
-    //   out::error() << "Cannot resize if not owning the data.\n";
-    // }
+
     if (new_n_size > n_capacity_) {
       out::error() << "Trying to resize vector to " << new_n_size 
                    << " elements but memory allocated only for " << n_capacity_ << "\n";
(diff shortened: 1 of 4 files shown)


resolve/vector/Vector.cpp
```diff
@@ -55,32 +44,38 @@ namespace ReSolve { namespace vector {
...
+    return n_capacity_;
   }
 
   /** 
-   * @brief get the current number of elements in a single vector 
-   * (use only for vectors with changing sizes, allocate for maximum expected size).
+   * @brief get the number of elements in a single vector.
+   * 
+   * For vectors with changing sizes, set the vector capacity to
+   * maximum expected size.
```
shakedregev, line 63:
```suggestion
   * the maximum expected size.
```

PR #276, opened 2025-05-20
Fix compiler warnings in matrix handlers

pelesh:
## Description
 
Fix warnings in matrix handlers, mainly not using `override` for methods in derived classes.
 

 ## Proposed changes
 
Straightforward fix.
 
 ## Checklist
 
 
- [x] All tests pass. Code tested on
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
- [x] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [x] The new code follows Re::Solve style guidelines.
- N/A There are unit tests for the new code.
- N/A The new code is documented.
- [x] The feature branch is rebased with respect to the target branch.
 
 
 ## Further comments

PR #278, opened 2025-05-21
Track updated status for each vector in multivector (second attempt)

### LINKED ISSUE #216: Vector initialization functions do not set "updated" flags correctly

### LINKED ISSUE #259: `Vector::getData` function incorrectly syncs memory spaces.

pelesh:
## Description
 
 ## Proposed changes
 
This PR provides following:
- Separate bookkeeping of host and device memory updates for each vector in the multivector object.
- Add `syncData` overload to sync only one vector in multivector object. Closes #259.

 ## Checklist
 
 
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 
 ## Further comments
 
 This is still a draft. #277, #276 and #275 need to be merged before this PR is ready.

### DIFF
--- a/resolve/MemoryUtils.hpp
+++ b/resolve/MemoryUtils.hpp
@@ -64,10 +64,26 @@ namespace ReSolve
       ///
 
       template <typename I, typename T>
+      int allocateArrayOnHost(T** v, I n)
+      {
+        std::size_t arraysize = static_cast<std::size_t>(n) * sizeof(T);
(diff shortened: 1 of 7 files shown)


pelesh:
Close in favor of #314.

PR #294, opened 2025-06-02
Unit testing for Sparse matrix class

### LINKED ISSUE #280: Write unit tests for Sparse (matrix) class methods
### Summary

[Sparse (matrix) class](https://github.com/ORNL/ReSolve/blob/develop/resolve/matrix/Sparse.cpp) has no unit tests. We should create unit tests so we are confident they are not introducing bugs or leaks in the code.

### Rationale

There is no way to validate methods in this class.

### Description

_No response_

### Additional information

_No response_

adhamsi:
## Description
 
 Implements unit testing for `Sparse.cpp`.
 
 Closes #280
 
 ## Proposed changes
 
 Tests implemented:
- constructor
- setting data pointers
- setting values pointer
- copying values
- copying values, then attempting to set data/values (must result in error)
- allocating and destroying memory
 ## Checklist
 
 merging your code._
 
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 
 ## Further comments

### DIFF
--- a/tests/unit/matrix/runSparseTests.cpp
+++ b/tests/unit/matrix/runSparseTests.cpp
@@ -26,17 +26,24 @@ void runTests(const std::string& backend, ReSolve::memory::MemorySpace memspace,
   ReSolve::io::Logger::setVerbosity(ReSolve::io::Logger::NONE);
 
   result += test.constructor(50, 50, 2);
+  result += test.constructor(50, 100, 2);
   result += test.constructor(50, 50, 2, true);
   result += test.constructor(50, 50, 2, false, false);
 
   result += test.setDataPointers(50, 50, 2);
+  result += test.setDataPointers(50, 100, 2);
   result += test.setValuesPointer(50, 50, 2);
+  result += test.setValuesPointer(50, 100, 2);
 
   result += test.copyValues(50, 50, 2);
+  result += test.copyValues(50, 100, 2);
   result += test.copyValuesAndSetValues(50, 50, 2);
+  result += test.copyValuesAndSetValues(50, 100, 2);
   result += test.copyValuesAndSetDataPointers(50, 50, 2);
+  result += test.copyValuesAndSetDataPointers(50, 100, 2);
 
   result += test.allocateAndDestroyData(50, 50, 2);
+  result += test.allocateAndDestroyData(50, 100, 2);
 }
 
 int main(int, char**)


shakedregev (changes requested):
Tests are spuriously passing on Frontier (likely due to some pointer comparison that happens to be true). They are failing on CUDA, almost certainly because of the code I pointed out (and other places where it's repeated).

tests/unit/matrix/SparseTests.hpp
```diff
@@ -0,0 +1,344 @@
...
+     * checks if the pointers are set correctly.
+     * 
+     * @param n - number of rows
+     * @param m - number of columns
+     * @param nnz - number of non-zeros
+     * @return TestOutcome indicating success or failure of the test
+     */
+    TestOutcome setDataPointers(index_type n, index_type m, index_type nnz)
+    {
+      assert(nnz % m == 0 && "For this test, nnz must be divisible by m");
```
shakedregev, line 78: I think `nnz % m == 0` as well

tests/unit/matrix/SparseTests.hpp
```diff
@@ -0,0 +1,346 @@
...
+     * 
+     * @param n - number of rows
+     * @param m - number of columns
+     * @param nnz - number of non-zeros
+     * @return TestOutcome indicating success or failure of the test
+     */
+    TestOutcome setDataPointers(index_type n, index_type m, index_type nnz)
+    {
+      assert(nnz % n == 0 && "For this test, nnz must be divisible by m");
+
```
pelesh, line 79: If you need an assert like this, there is probably something wrong with the function. The message in the assertion is incorrect, too. Consider inputs like number of rows and row density.

tests/unit/matrix/SparseTests.hpp
```diff
@@ -0,0 +1,346 @@
...
+  {
+  public:
+    SparseTests(ReSolve::MatrixHandler& handler) : handler_(handler)
+    {
+      if (handler_.getIsCudaEnabled() || handler_.getIsHipEnabled()) {
+        memspace_ = memory::DEVICE;
+      } else {
+        memspace_ = memory::HOST;
+      }
+    }
```
pelesh, line 25: Why create matrix handler? These are just matrix class tests.

tests/unit/matrix/SparseTests.hpp
```diff
@@ -0,0 +1,346 @@
...
+     * @param m - number of columns
+     * @param nnz - number of non-zeros
+     * @return TestOutcome indicating success or failure of the test
+     */
+    TestOutcome setDataPointers(index_type n, index_type m, index_type nnz)
+    {
+      assert(nnz % n == 0 && "For this test, nnz must be divisible by m");
+      
+      TestStatus status;
+      status = true;
```
pelesh, line 81: Consider "success" as the variable name rather than "status". This makes intent of the code more clear.
```suggestion
      TestStatus success;
      success = true;
```

tests/unit/matrix/SparseTests.hpp
```diff
@@ -0,0 +1,346 @@
...
+    {
+      TestStatus status;
+      status = true;
+
+      ReSolve::matrix::Csr* A = new ReSolve::matrix::Csr(n, m, nnz);
+
+      real_type* val_data = new real_type[nnz];
+      for (index_type i = 0; i < nnz; ++i) {
+        val_data[i] = static_cast<real_type>(i + 1);
+      }
```
pelesh, line 281: Create a helper function to check operation result.

tests/unit/matrix/SparseTests.hpp
```diff
@@ -0,0 +1,346 @@
...
+     * @brief Verify that matrix will not reset values pointer when it currently owns data.
+     *
+     * Copies a data array using `copyValues`, then attempts to set the values pointer, 
+     * and verifies that this results in an error.
+     *
+     * @return TestOutcome indicating success or failure of the test
+     */
+    TestOutcome copyValuesAndSetValues(index_type n, index_type m, index_type nnz)
+    {
+      TestStatus status;
```
pelesh, line 273: Create separate tests for copy and set matrix values.

tests/unit/matrix/SparseTests.hpp
```diff
@@ -0,0 +1,346 @@
...
+            mem_.copyArrayDeviceToHost(h_val_data, A->getValues(memory::DEVICE), nnz);
+          }
+
+          // Check if the copied values are correct
+          for (index_type i = 0; i < nnz; ++i) {
+            if (h_val_data[i] != static_cast<real_type>(i + 1)) {
+              std::cout << "Copied values do not match expected values.\n";
+              status = false;
+              break;
+            }
```
pelesh, line 195: Create a helper function to check results and reuse it across relevant tests.

pelesh (changes requested):
This is a good start but requires some more work. Setting this as a draft.

tests/unit/matrix/SparseTests.hpp
```diff
@@ -0,0 +1,346 @@
...
+     * Copies values into the sparse matrix and then attempts to set the values
+     * pointer, which should fail since the matrix owns the data.
+     * 
+     * @param n - number of rows
+     * @param m - number of columns
+     * @param nnz - number of non-zeros
+     * @return TestOutcome indicating success or failure of the test
+     */
+    TestOutcome copyValuesAndSetDataPointers(index_type n, index_type m, index_type nnz)
+    {
```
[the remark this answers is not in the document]
shakedregev: The test is probably unnecessary. If you rebased with respect to develop, this should throw a segfault. It doesn't really make sense to have a test that passes only if there's a segfault.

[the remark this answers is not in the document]
adham-ibrahim7: I'm up to date with respect to develop. Which commit are you referring to?

tests/unit/matrix/SparseTests.hpp
pelesh: Since you don't want to use the entire `MatrixHandler` object only for this, consider something like:
```c++
#ifdef RESOLVE_USE_GPU
memspace_ = memory::DEVICE
#else
memspace_ = memory::HOST
#endif
```

tests/unit/matrix/SparseTests.hpp
```diff
@@ -0,0 +1,361 @@
...
+
+      real_type* val_data = new real_type[nnz];
+      for (index_type i = 0; i < nnz; ++i) {
+        val_data[i] = static_cast<real_type>(i + 1);
+      }
+
+      if (A->copyValues(val_data, memory::HOST, memspace_) != 0) {
+        std::cout << "Failed to copy values.\n";
+        success = false;
+      }
```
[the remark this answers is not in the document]
adham-ibrahim7: We've tested that in a different test, so I thought it wouldn't be necessary here.

shakedregev (commented):
Looks good and tests pass. Just add tests for rectangular matrices and let's make sure they still pass.

pelesh (approved):

PR #305, opened 2025-06-09
Update redundant scaling kernel names

### LINKED ISSUE #301: Scaling kernels and functions are redundantly named
### Summary

1. Remove the word "Diag" or "Diagonal" where "Scale" is mentioned.
2. Remove reference to vectors or matrices within their namespace. For example`vectorDiagonalScale` becomes `vector::Scale`.

### Rationale

1. We want to remove redundancy by mentioning Diagonal when scaling, because all scaling is diagonal. 
2. Remove redundant references to vectors or matrices and put them within their namespaces.

### Description

These changes should be applied to `leftDiagScale`, `rightDiagScale` and `vectorDiagScale` along with all functions in the stack that call them.

### Additional information

_No response_

adham-ibrahim7:
## Description
 
Renames the functions to scale a vector by a diagonal matrix, and moves this functionality to `VectorHandler.cpp` rather than `MatrixHandler.cpp`.

Closes #301
 

 ## Proposed changes
 
- Renames `leftDiagonalScale` to `leftScale` and similarly for `rightDiagonalScale` and `vectorScale` in all declarations.
- Moves the `vectorScale` function from `MatrixHandler.hpp` and all derived classes to `VectorHandler.hpp` and its derived classes, as well as moving the test for it.

 
 ## Checklist
 
 merging your code._
 
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 
 ## Further comments

### DIFF
--- a/resolve/cuda/cudaVectorKernels.cu
+++ b/resolve/cuda/cudaVectorKernels.cu
@@ -58,7 +58,7 @@ namespace ReSolve
 
     } // namespace kernels
 
-    void cuda_set_array_const(index_type n, real_type val, real_type* arr)
+    void setArrayConst(index_type n, real_type val, real_type* arr)
     {
       index_type num_blocks;
       index_type block_size = 512;
@@ -66,7 +66,7 @@ namespace ReSolve
       kernels::set_const<<<num_blocks, block_size>>>(n, val, arr);
     }
 
-    void cudaAddConst(index_type n, real_type val, real_type* arr)
+    void addConst(index_type n, real_type val, real_type* arr)
     {
       index_type num_blocks;
       index_type block_size = 512;
--- a/resolve/cuda/cudaVectorKernels.cu
+++ b/resolve/cuda/cudaVectorKernels.cu
@@ -10,66 +10,68 @@
 #include <cuda_runtime.h>
 
 #include <resolve/cuda/cudaVectorKernels.h>
-
+#include <resolve/cuda/cudaKernels.h>
 
 namespace ReSolve
 {
-  namespace kernels
-  {
-
-    /**
-     * @brief CUDA kernel that sets values of an array to a constant.
-     *
-     * @param[in]  n   - length of the array
-     * @param[in]  val - the value the array is set to
-     * @param[out] arr - a pointer to the array
-     *
-     * @pre  `arr` is allocated to size `n`
-     * @post `arr` elements are set to `val`
-     */
-    __global__ void set_const(index_type n, real_type val, real_type* arr)
+  namespace cuda {
+    namespace kernels
     {
-      index_type i = blockIdx.x * blockDim.x + threadIdx.x;
-      if(i < n)
+
+      /**
+       * @brief CUDA kernel that sets values of an array to a constant.
+       *
+       * @param[in]  n   - length of the array
+       * @param[in]  val - the value the array is set to
+       * @param[out] arr - a pointer to the array
+       *
+       * @pre  `arr` is allocated to size `n`
+       * @post `arr` elements are set to `val`
+       */
+      __global__ void set_const(index_type n, real_type val, real_type* arr)
       {
-        arr[i] = val;
+        index_type i = blockIdx.x * blockDim.x + threadIdx.x;
+        if(i < n)
+        {
+          arr[i] = val;
+        }
       }
+
+      /**
+       * @brief CUDA kernel that adds a constant to each element of an array.
+       *
+       * @param[in]       n   - length of the array
+       * @param[in]       val - the value to add to each element
+       * @param[in, out]  arr - a pointer to the array
+       *
+       * @pre  `arr` is allocated to size `n`
+       * @post `val` is added to each element of `arr`
+       */
+      __global__ void addConst(index_type n, real_type val, real_type* arr)
+      {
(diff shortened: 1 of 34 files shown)


adham-ibrahim7:
@shakedregev You suggested removing vector from the name of the function and leaving it as just `Scale`, but this would conflict with the existing `VectorHandler:scal(const real_type* alpha, vector::Vector* x, memory::MemorySpace memspace)` which scales by a constant. I left it as `vectorScale` in this namespace, but if you have a suggestion for a different name let me know. The names in `VectorHandler` should all probably be updated, they don't follow convention and are very abbreviated.

shakedregev (commented):
Make the change to the test and wait for approval from Slaven on the new naming scheme. Otherwise looks good.

pelesh:
> @shakedregev You suggested removing vector from the name of the function and leaving it as just `Scale`, but this would conflict with the existing `VectorHandler:scal(const real_type* alpha, vector::Vector* x, memory::MemorySpace memspace)` which scales by a constant. I left it as `vectorScale` in this namespace, but if you have a suggestion for a different name let me know. The names in `VectorHandler` should all probably be updated, they don't follow convention and are very abbreviated.

I also think simply `scale` would be a way to go. I prefer that.It is a function overload, so it would not conflict with scale by parameter.

pelesh:
Since you are already at it, I would use namespaces to distinguish between kernels, wrappers and handler functions:
- Use `scale` as vector handler method name.
- Use e.g. `cuda::scale` for the wrapper.
- Use e.g. `cuda::kernel::scalę` for the kernel.

adham-ibrahim7:
> > @shakedregev You suggested removing vector from the name of the function and leaving it as just `Scale`, but this would conflict with the existing `VectorHandler:scal(const real_type* alpha, vector::Vector* x, memory::MemorySpace memspace)` which scales by a constant. I left it as `vectorScale` in this namespace, but if you have a suggestion for a different name let me know. The names in `VectorHandler` should all probably be updated, they don't follow convention and are very abbreviated.
> 
> I also think simply `scale` would be a way to go. I prefer that.It is a function overload, so it would not conflict with scale by parameter.

It may be confusing to a user that scaling by a constant is `scal` (without the "e") and element wise multiplication is `scale`. I can go ahead with renaming to just `scale` if you think its best.

shakedregev:
> > > @shakedregev You suggested removing vector from the name of the function and leaving it as just `Scale`, but this would conflict with the existing `VectorHandler:scal(const real_type* alpha, vector::Vector* x, memory::MemorySpace memspace)` which scales by a constant. I left it as `vectorScale` in this namespace, but if you have a suggestion for a different name let me know. The names in `VectorHandler` should all probably be updated, they don't follow convention and are very abbreviated.
> > 
> > 
> > I also think simply `scale` would be a way to go. I prefer that.It is a function overload, so it would not conflict with scale by parameter.
> 
> It may be confusing to a user that scaling by a constant is `scal` (without the "e") and element wise multiplication is `scale`. I can go ahead with renaming to just `scale` if you think its best.

See Slaven's comment a few lines above. They should both be "scale", but if you supply a scalar, it basically assumes it's a constant vector. If you supply a vector, then it's element-wise. The two functions accept different arguments and are therefore different to the compiler. The user gets the correct version as long as they supply the correct input.

resolve/cuda/cudaVectorKernels.cu
```diff
@@ -10,66 +10,111 @@
...
+      {
+        index_type i = blockIdx.x * blockDim.x + threadIdx.x;
+        if(i < n)
+        {
+          arr[i] += val;
+        }
+      }
+
+      /**
+       * @brief Scales a vector by a diagonal matrix
```
shakedregev, line 60:
```suggestion
       * @brief Scales a vector by a diagonal matrix represented by a vector
```

resolve/cuda/cudaVectorKernels.cu
```diff
@@ -10,66 +10,111 @@
...
+    {
+      index_type num_blocks;
+      index_type block_size = 512;
+      num_blocks = (n + block_size - 1) / block_size;
+      kernels::addConst<<<num_blocks, block_size>>>(n, val, arr);
     }
 
     /**
-     * @brief CUDA kernel that adds a constant to each element of an array.
+     * @brief Wrapper that scales a vector by a diagonal matrix
```
shakedregev, line 101:
```suggestion
     * @brief Wrapper that scales a vector by a diagonal matrix represented by a vector
```

resolve/hip/hipVectorKernels.hip
```diff
@@ -15,53 +15,96 @@
...
-     *
-     * @pre  `arr` is allocated to size `n`
-     * @post `val` is added to each element of `arr`
-     */
-    __global__ void addConst(index_type n, real_type val, real_type* arr)
-    {
-      index_type i = blockIdx.x * blockDim.x + threadIdx.x;
-      if(i < n)
+      /**
+      * @brief Scales a vector by a diagonal matrix
```
shakedregev, line 51:
```suggestion
      * @brief Scales a vector by a diagonal matrix represented by a vector
```

shakedregev (approved):
Good job! Approving subject to retesting. Will merge after.

pelesh (changes requested):
Please wait for #307 to be merged, first.

PR #307, opened 2025-06-12
Fix memory leaks

pelesh:
## Description
 
Address sanitizer identified several memory leaks.
 

 ## Proposed changes
 
Fix memory leaks on CPU hardware first.

In FGMRES classes, rather than deleting vector views created on heap, create them on stack instead.
 
 ## Checklist
 
 
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
- N/A There are unit tests for the new code.
- N/A The new code is documented.
 
 
 ## Further comments


pelesh (commented):
Looks good, one minor correction needed.

I'd still want to test `GramSchmidt` class and random sketching a little more.

resolve/LinSolverIterativeRandFGMRES.cpp
```diff
@@ -350,11 +344,11 @@ namespace ReSolve
         if (sketching_method_ == cs) {
           vec_S_->setToZero(memspace_);
         }
-        vec_v->setData(vec_V_->getVectorData(0, memspace_), memspace_);
-        vec_s->setData(vec_S_->getVectorData(0, memspace_), memspace_);
-        sketching_handler_->Theta(vec_v, vec_s);
+        vec_v.setData(vec_V_->getVectorData(0, memspace_), memspace_);
+        vec_s.setData(vec_S_->getVectorData(0, memspace_), memspace_);
+        sketching_handler_->Theta(&vec_v, &vec_s);
```
shakedregev, line 349: Function name should start with lower case letter.

resolve/matrix/Csr.cpp
```diff
@@ -73,6 +73,7 @@ namespace ReSolve
         h_val_data_ = *vals;
         h_data_updated_ = true;
         owns_cpu_sparsity_pattern_ = true;
+        owns_cpu_values_ = true;
```
shakedregev, line 76: The cases here are not listed sequentially. This makes it harder to follow.

superwhiskers (approved):
to be fair i wrote a lot of this but i don't see anything wrong with the changes here

shakedregev (approved):
Tests pass and I'm approving. See minor comments.

PR #314, opened 2025-06-14
Track updated status for each vector in multivector (third attempt)

### LINKED ISSUE #216: Vector initialization functions do not set "updated" flags correctly

### LINKED ISSUE #259: `Vector::getData` function incorrectly syncs memory spaces.

### LINKED ISSUE #313: Remove VectorHandler from VectorTests
### Summary

Replace `VectorHandler` instance in `VectorTests` class.

### Rationale

`VectorTests` implements unit tests for `vector::Vector` class methods and should not depend on `VectorHandler`, which is tested in separate unit test.

### Description

_No response_

### Additional information

_No response_

pelesh:
## Description
 
@superwhiskers 

 ## Proposed changes
 
This PR provides following:
- Separate bookkeeping of host and device memory updates for each vector in the multivector object.
- Add `syncData` overload to sync only one vector in multivector object. Closes #259.
- Fix bug in `GramSchmidt` class where multi vector view was set to a wrong size.
- Remove duplicate `getVectorData` function.
- Closes #313.
 
 ## Checklist
 
 
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 
 ## Further comments
 
This PR supersede #278.

### DIFF
--- a/resolve/vector/Vector.cpp
+++ b/resolve/vector/Vector.cpp
@@ -139,7 +139,6 @@ namespace ReSolve { namespace vector {
    * Use this function if you update vector elements by accessing the raw data
    * pointer.
    *
-   *
    * @param[in] memspace - Memory space (HOST or DEVICE)
    * 
    * @warning This is an expert level method. Use only if you know what


tests/unit/vector/runVectorTests.cpp
```diff
@@ -3,57 +3,51 @@
...
+    result += test.setToConst(50);
+    result += test.setToConst(50);
+  }
 
-#ifdef RESOLVE_USE_HIP
-  runTests<ReSolve::LinAlgWorkspaceHIP>("HIP", result);
+#ifdef RESOLVE_USE_GPU
+  {
+    ReSolve::tests::VectorTests test(ReSolve::memory::DEVICE);
+
```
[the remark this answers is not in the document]
pelesh: These are used to define `test` variable in a local scope.

resolve/GramSchmidt.cpp
```diff
@@ -93,6 +93,7 @@ namespace ReSolve
 
     vec_w_ = new vector_type(n);
     vec_v_ = new vector_type(n);
```
kswirydo, line 95: Shouldn't v be multivector if "variant" is either low-synch or pm_mgs?

kswirydo: okay I see you introduced "x"

resolve/GramSchmidt.hpp
```diff
@@ -48,6 +48,7 @@ namespace ReSolve
 
       vector_type* vec_v_{nullptr}; // aux variable
       vector_type* vec_w_{nullptr}; // aux variable
+      vector_type* vec_x_{nullptr}; // aux variable
```
kswirydo, line 51: Would suggest using something like `vec_y` or `vec_a`. `x` is typically reserved for solution vector. Not that this would be mixed up with GMRES/other solvers but might become hard to work on/read/understand in the future.

resolve/vector/Vector.hpp
```diff
@@ -33,33 +34,37 @@ namespace ReSolve { namespace vector {
       int copyDataFrom(const real_type* data, memory::MemorySpace memspaceIn, memory::MemorySpace memspaceOut);
       int copyDataFrom(Vector* v, memory::MemorySpace memspaceIn, memory::MemorySpace memspaceOut);
       real_type* getData(memory::MemorySpace memspace);
-      real_type* getData(index_type i, memory::MemorySpace memspace); // get pointer to i-th vector in multivector
+      real_type* getData(index_type i, memory::MemorySpace memspace);
```
kswirydo, line 37: what was wrong with this comment :D

resolve/vector/VectorHandlerHip.cpp
```diff
@@ -234,7 +234,7 @@ namespace ReSolve {
     using namespace constants;
 
     if (k < 200) {
-      hip::mass_inner_product_two_vectors(size, k, x->getData(memory::DEVICE) , x->getData(1, memory::DEVICE), V->getData(memory::DEVICE), res->getData(memory::DEVICE));
+      hip::mass_inner_product_two_vectors(size, k, x->getData(0, memory::DEVICE) , x->getData(1, memory::DEVICE), V->getData(memory::DEVICE), res->getData(memory::DEVICE));
```
kswirydo, line 237: Code formatting is totally wrong here.

tests/unit/vector/VectorTests.hpp
```diff
@@ -19,14 +19,9 @@ namespace ReSolve {
     class VectorTests : TestBase
     {
       public:
-        VectorTests(ReSolve::VectorHandler& handler) : handler_(handler)
+        VectorTests(memory::MemorySpace memspace = memory::HOST)
+          : memspace_(memspace)
         {
```
kswirydo, line 24: Why was this removed (confused emoji)

tests/unit/vector/runVectorTests.cpp
```diff
@@ -3,57 +3,51 @@
...
-/**
- * @brief Run vector tests with a given backend
- *
- * @tparam WorkspaceType workspace type LinAlgWorkspace{Cpu, CUDA, HIP} supported
- * @param[in] backend - name of the hardware backend
- * @param[out] result - test results
- */
-template<typename WorkspaceType>
-void runTests(const std::string& backend, ReSolve::tests::TestingResults& result)
-{
```
kswirydo, line 15: Why is the code below removed?

kswirydo (commented):
See comments

tests/unit/vector/runVectorTests.cpp
pelesh: Vector class does not need the workspace. This is just removing redundant dependencies from the unit test.

PR #320, opened 2025-06-25
fix some more memory leaks and add ubsan support to `CMakeLists.txt`

superwhiskers:
## description

fixes a memory leak introduced in `4a7c9847e32d6dade0d07e61db62e8f93dcbf58e` and another in `afdd00f4f2e9bc851daca48191d543e3a0ee5d1b`. adds ubsan support and sets the asan flags in cflags and fortran flags just in case (i didn't find anything by doing this but it may in future code iterations).

 ## checklist

     - [x] CPU backend
     - [ ] CUDA backend
     - [ ] HIP backend


shakedregev (approved):
Looks good and tests pass. Merging.

PR #333, opened 2025-07-11
HyKKT GPU Permutation Kernels (Attempt 2)

### LINKED ISSUE #309: Permutation class and tests for HyKKT
### Summary

Port [this test](https://github.com/ORNL/hykkt/blob/develop/examples/perm_driver.cpp) to ReSolve and implement the necessary classes and functions on cpu/cuda/hip (or copy implementations as possible).

### Rationale

Taking a dense permutation and turning it into a sparse one is necessary for HyKKT (and potentially for other solvers).

### Description

1. Any originally cpu functions need only have cpu implementations
2. Any originally gpu functions should have cpu/cuda/hip support.
3. Write tests as within the rest of ReSolve.
4. Add tests for square and rectangular matrices.

### Additional information

_No response_

adhamsi:
## Description
 
Implements the GPU kernels for the Permutation class. The `Permutation` class contains the following:
- `addMatrixInfo` to add the Hessian, Jacobian, and transpose the Jacobian
- both `symAmd` and `addCustomPerm` as options to configure the permutation before the first iteration
- `invertPerm`, performed once in the first iteration
- `vecMapRC`, `vecMapR`, `vecMapC`, performed once in the first iteration
- `mapIdx` to map values arrays in subsequent iterations of the solver
 
Closes #309.

 ## Proposed changes
 
This builds on a previous attempt that performed operations on device that must be done on host. Now, the class functions assume that the matrices passed in are updated on host, and performs all setup functions on host. Then, the `mapIdx` function is called on whatever memory space is passed to the constructor initially. This requires storing both host and device copies of the computed permuation arrays in the class.
 
 ## Checklist
 
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 
 ## Further comments

### DIFF
--- a/resolve/hykkt/permutation/Permutation.cpp
+++ b/resolve/hykkt/permutation/Permutation.cpp
@@ -243,8 +243,8 @@ namespace ReSolve
      * @post new_val contains the permuted old_val
      */
     void Permutation::mapIndex(PermutationType permutation,
-                               real_type*         old_val,
-                               real_type*         new_val)
+                               real_type*      old_val,
+                               real_type*      new_val)
     {
       index_type  length;
       index_type* apply_perm_;


resolve/hykkt/permutation/Permutation.hpp
```diff
@@ -0,0 +1,108 @@
...
+  namespace hykkt
+  {
+    enum PermutationType
+    {
+      PERM_V,
+      REV_PERM_V,
+      PERM_HES_V,
+      PERM_JAC_V,
+      PERM_JAC_TR_V
+    };
```
superwhiskers, line 36: it may be more appropriate to use a scoped enumeration (an `enum class`) here

superwhiskers (commented):
since this is a draft right now, i'm just going to leave this as a comment

Issue #338, opened 2025-07-18
Add methods for extracting F Factor from KLU


shakedregev:
### Summary

Add a method to return F from the KLU $`A=LU+F`$ factorization.

### Rationale

Currently ReSolve completely ignores the existence of F in the $`A=LU+F`$ factorization. This means that for matrices where F is substantial, the user might get wrong results when using refactorization on the gpu, with no method for debugging.

### Description

Have a method that is analogous for the ones to get L and U with no extra work.

Issue #332, opened 2025-07-11
Correct usage of Re::Solve?


pelesh:
> My application is to factorize `A` and then solve `Ax = b` for various `b` and sometimes update  the `A` without changing its sparsity pattern (only the scaling i.e. the diagonal is updated). This is how I do it for now:
> 
> ```c
> int __initialize_work(LinSysWork *work)
>   {
>     work->workspace_hip = new ReSolve::LinAlgWorkspaceHIP;
>     work->workspace_hip->initializeHandles();
>     work->solver = new ReSolve::SystemSolver(work->workspace_hip,
>                                   "klu", // factorization
>                                   "rocsolverrf", // refactorization
>                                   "rocsolverrf", // triangular solve
>                                   "none", // preconditioner (always 'none' here)
>                                   "none"); // iterative refinement
> 
>     // initialize ReSolve members of work:
>     int nnz = work->kkt->p[work->kkt->n]; // The last element of A->work gives the number of non-zeros
>     // The work->KKT matrix is CSC (lower-triangular is stored only)
>     // by symmetry this is CSR (upper-triangular is stored only)
>     // symmetric and not expanded
>     work->mat_A = new ReSolve::matrix::Csr(
>       work->kkt->n, work->kkt->m, nnz, true, false);
>     // work->mat_A->setDataPointers(
>     //   work->kkt->p, work->kkt->i, work->kkt->x, ReSolve::memory::HOST);
> 
>     work->mat_A->allocateMatrixData(ReSolve::memory::HOST);
>     work->mat_A->copyDataFrom(
>       work->kkt->p, work->kkt->i, work->kkt->x, ReSolve::memory::HOST,
>       ReSolve::memory::HOST);
>     work->mat_A->allocateMatrixData(ReSolve::memory::DEVICE);
>     work->mat_A->syncData(ReSolve::memory::DEVICE);
> 
>     work->mat_A->print(std::cout, 1);
> 
>     work->vec_rhs = new ReSolve::vector::Vector(work->mat_A->getNumRows());
>     work->vec_rhs->allocate(ReSolve::memory::DEVICE);
>     work->vec_rhs->allocate(ReSolve::memory::HOST);
>     work->vec_rhs->setToConst(1.0, ReSolve::memory::DEVICE);
>     work->vec_rhs->syncData(ReSolve::memory::HOST);
> 
>     work->vec_x = new ReSolve::vector::Vector(work->mat_A->getNumColumns());
[...]

PR #339, opened 2025-07-18
added methods to extract F Factors

shakedregev:
## Description
 
 _ReSolve was completely ignoring the existence of F in the KLU factorization. The user will now have methods to examine F to at least see if ignoring F is valid mathematically._
 
 _Closes #338. Is necessary for solving #332  _

 ## Proposed changes
 
 The changes allow the user to debug, verify the existence and magnitude of nonzero entries in F, and use F in an external solver should a method for this arise._
 
 ## Checklist
 
 merging your code._
 
- [ ] All tests pass. Code tested on
     - [ ] CPU backend
     - [ ] CUDA backend
     - [ ] HIP backend

### DIFF
--- a/resolve/LinSolverDirectKLU.cpp
+++ b/resolve/LinSolverDirectKLU.cpp
@@ -369,13 +369,17 @@ namespace ReSolve
   {
     if (!factors_extracted_)
     {
+      const int n = Symbolic_->n;
       const int nnzL = Numeric_->lnz;
       const int nnzU = Numeric_->unz;
+      const int nnzF = Numeric_->Offp [n];
 
       L_ = new matrix::Csc(A_->getNumRows(), A_->getNumColumns(), nnzL);
       U_ = new matrix::Csc(A_->getNumRows(), A_->getNumColumns(), nnzU);
+      F_ = new matrix::Csc(A_->getNumRows(), A_->getNumColumns(), nnzF);
       L_->allocateMatrixData(memory::HOST);
       U_->allocateMatrixData(memory::HOST);
+      F_->allocateMatrixData(memory::HOST);
 
       int ok = klu_extract(Numeric_,
                            Symbolic_,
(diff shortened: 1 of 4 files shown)


shakedregev:
There aren't other solvers that use F. This will allow at least to examine the factor F and see if ignoring it is reasonable.

resolve/LinSolverDirectKLU.cpp
```diff
@@ -411,13 +416,18 @@ namespace ReSolve
...
     if (!factors_extracted_)
     {
+      const int n    = Symbolic_->n;
       const int nnzL = Numeric_->lnz;
       const int nnzU = Numeric_->unz;
+      const int nnzF = Numeric_->Offp[n];
 
       L_ = new matrix::Csc(A_->getNumRows(), A_->getNumColumns(), nnzL);
       U_ = new matrix::Csc(A_->getNumRows(), A_->getNumColumns(), nnzU);
+      F_ = new matrix::Csc(A_->getNumRows(), A_->getNumColumns(), nnzF);
```
superwhiskers, line 426: add code to free `F_` in the destructor. this will get leaked otherwise

superwhiskers (changes requested):
possible memory leak. also, it may be a good idea to add a test that involves the `F` factor

also, the thing i mentioned above about other ways of handling factor extraction if we ever add more than $A = LU$ and $A = LU + F$ factorization. that's also something to consider.

otherwise, this pull request looks fine

resolve/LinSolverDirectKLU.cpp
```diff
@@ -369,13 +371,17 @@ namespace ReSolve
   {
     if (!factors_extracted_)
     {
+      const int n    = Symbolic_->n;
```
superwhiskers, line 374: we should probably add a method to extract the factors if they aren't already at this point, with the amount of code required to do that. this will avoid duplicating this code three times.

superwhiskers (changes requested):

Issue #337, opened 2025-07-18
Bug with transpose after memory synchronization update


adhamsi:
`adham/ruiz-transpose-issue` has been rebased with the new commit on develop that adds the memory synchronization on HIP. Changing the test in `runHykktRuizScalingTests.cpp` from 64 to 256 shows bad output for the transpose. It is outputting `J_tr->getRowData(memspace_)`. 

The synchronization fixed the issue when the test cases are 8 and 12, but the behavior is still erratic. The only value I've found that it breaks for is 256. There must be another issue.

superwhiskers:
this is because we [don't resize the transpose buffer](https://github.com/ORNL/ReSolve/blob/develop/resolve/matrix/MatrixHandlerHip.cpp#L267-L281) when it's necessary. a similar issue is likely the culprit for cuda because [we do the same thing there too](https://github.com/ORNL/ReSolve/blob/6e3c6ec2e1e5e1a791173cbcc0673ffd601578f6/resolve/matrix/MatrixHandlerCuda.cpp#L294-L315). a dumb fix for this would be to always allocate a new transpose buffer.

PR #343, opened 2025-07-22
Added the ability to reset a workspace without completely destroying it.

shakedregev:
## Description
 
 _Currently our tests reuse the same handler and workspace without resetting the buffers. This causes problems when doing multiple tests of different sizes._
 
 _Closes #337_
_Closes #9_ 
 
 ## Proposed changes
 
 _I added methods to reset the buffers and used them in the matrix handler tests. Other tests might also need to reuse them._
 
 ## Checklist
 
 merging your code._
 
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend

### DIFF
--- a/resolve/workspace/LinAlgWorkspaceCpu.cpp
+++ b/resolve/workspace/LinAlgWorkspaceCpu.cpp
@@ -19,5 +19,6 @@ namespace ReSolve
   void LinAlgWorkspaceCpu::resetLinAlgWorkspace()
   {
     // No resources to reset in CPU workspace
+    return;
   }
 } // namespace ReSolve


resolve/workspace/LinAlgWorkspaceCpu.cpp
```diff
@@ -15,4 +15,9 @@ namespace ReSolve
   void LinAlgWorkspaceCpu::initializeHandles()
   {
   }
+
+  void LinAlgWorkspaceCpu::resetLinAlgWorkspace()
+  {
+    // No resources to reset in CPU workspace
```
shakedregev, line 21:
```suggestion
    // No resources to reset in CPU workspace
    return;
```

superwhiskers (approved):
this both appears to work cleanly and fixes the issue brought up within #337

Issue #341, opened 2025-07-18
Make KLU not create an F factor when using other solvers.


shakedregev:
### Summary

Force KLU to do a true $`A=LU`$ factorization.

### Rationale

Other solvers do not use an F factor, but we should not ignore it.
More details [here](https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/988).

### Description

Ideally we should let KLU run with its defaults when possible (when KLU is used for each iteration), but force it to not produce F when other solvers are used. This can be done with `Common->btf=0`.

PR #346, opened 2025-07-24
Enforce F being 0 and remove duplicate extract factors code

shakedregev:
## Description
 
 _We need to verify that F is 0 and simplify factor extraction._
 
 _Closes #341 _
 
 ## Proposed changes
 
This removes duplicate code, making it easier to edit factor extraction functions and asserts that F is 0, which is an underlying assumption in ReSolve.
 
 ## Checklist
 
 merging your code._
 
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend

### DIFF
--- a/resolve/LinSolverDirectKLU.cpp
+++ b/resolve/LinSolverDirectKLU.cpp
@@ -1,7 +1,7 @@
 #include "LinSolverDirectKLU.hpp"
 
 #include <cassert>
-#include <cstring> // includes memcpy
+#include <cstring>
 
 #include <resolve/matrix/Csc.hpp>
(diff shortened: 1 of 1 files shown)


resolve/LinSolverDirectKLU.cpp
```diff
@@ -1,5 +1,6 @@
 #include "LinSolverDirectKLU.hpp"
 
+#include <cassert> // includes assert
```
shakedregev, line 3:
```suggestion
#include <cassert>
```

superwhiskers (approved):
looks good to me

PR #350, opened 2025-07-29
Implement the HyKKT Cholesky module

### LINKED ISSUE #311: Add Cholesky factorization, triangular solves, and tests
### Summary

Write a different version of [this test](https://github.com/ORNL/hykkt/blob/develop/examples/cuSolver_driver_cholesky.cpp) within the ReSolve framework. Implement the necessary classes and functions on cpu/cuda/hip (or copy implementations as possible). Note that it is crucial to separate numeric and symbolic factorizations, because symbolic factorization is only called once. 

### Rationale

Cholesky factorization and solve are essential for HyKKT to work. A simpler version could be used for a single matrix in a series for other solvers for problems that are fully SPD.

### Description

1. Any originally cpu functions need only have cpu implementations
2. Any originally gpu functions should have cpu/cuda/hip support.
3. Write tests as within the rest of ReSolve.
4. Write a test that generates an SPD matrix of arbitrary (square) size, multiplies by a known solution x (with current matvec in ReSolve) to get the right hand side b, and then solves and verifies that the solution is correct. The simplest solution would be to test $`||Ax-b||/||b||`$

### Additional information

_No response_

adhamsi:
## Description
 
 This implements the Cholesky factorization module, one of the five HyKKT modules. The module supports computing the symbolic and numeric factorization of a sparse positive definite matrix (the H_gamma matrix from the HyKKT paper), and updating the nonzeros of the matrix and reusing the symbolic factorization. The module then provides a `solve(vector::Vector* x, vector::Vector* b)` function. 
 
Closes #311 

 ## Proposed changes
 
- The CUDA implementation is largely copied over from the original HyKKT repo and uses `cusolver`.
- The CPU implementation uses the suitesparse CHOLMOD module. This involves converting from the ReSolve CSR sparse type to the CHOLMOD-native `cholmod_sparse` type, which is an inideal solution, but acceptable given that the CPU implementation is primarily for correctness and comparison to GPU implementations.
- The HIP implementation uses the `rocsolver` library refactorization and triangular solve functions.
 
 ## Checklist
 
- [ ] All tests pass. Code tested on
     - [ ] CPU backend
     - [ ] CUDA backend
     - [ ] HIP backend


resolve/CMakeLists.txt
```diff
@@ -164,7 +164,6 @@ endif()
 if(RESOLVE_USE_KLU)
   add_subdirectory(hykkt)
 endif()
-#list(APPEND ReSolve_Targets_List resolve_hykkt)
```
shakedregev, line 167: Why are we removing this from the list of targets?

tests/unit/hykkt/HykktCholeskyTests.hpp
```diff
@@ -0,0 +1,327 @@
...
+
+      TestOutcome randomized(index_type n)
+      {
+        TestStatus  status;
+        std::string testname(__func__);
+        testname += " n = " + std::to_string(n);
+
+        cholmod_sparse* L      = randomSparseLowerTriangular((size_t) n);
+        cholmod_sparse* L_tr   = cholmod_transpose(L, 1, &Common);
+        cholmod_sparse* A_chol = cholmod_ssmult(L, L_tr, 0, 1, 0, &Common);
```
shakedregev, line 107: Misleading name choice. A_times_At would be better.

tests/unit/hykkt/HykktCholeskyTests.hpp
```diff
@@ -0,0 +1,327 @@
...
+
+        if (memspace_ == memory::DEVICE)
+        {
+          // x_expected->syncData(memory::HOST);
+          x->syncData(memory::HOST);
+        }
+
+        // Verify result
+        // TODO only works with 1e-4 tolerance
+        real_type tol = 1e-4;
```
shakedregev, line 142: We need to get this to around 10*machine_epsilon ($`~2.2e-15`$). Forcing a solution of all 1s, or a better conditioned matrix may help.

shakedregev (approved):
CUDA and CPU tests pass. Approved, pending documentation change requests. 
HIP is not working because as far as we can tell their Cholesky solver does not work (verified with @kswirydo from AMD). We will need to add a separate test for HIP once we figure out what is going on with their Cholesky solver.

kswirydo:
We are investigating the issue on AMD GPUs. I got the same failure for identity matrix too ...

PR #378, opened 2025-09-29
Issue # 376: Incorporating initial guess into residuals for iterative solvers

ASunil-17:
## Description
 
 The solution from the direct solve is treated as the initial guess. The iterative solvers are set up to compute an update. If the update plus the initial guess is offering a better estimation to the solution than the initial guess, then we return that quantity. If the initial residual is greater than the norm of the rhs, we assert that the initial guess is a poor choice. 
 
 _Closes # 376 
 
 _Mentions @shakedregev 
 

 ## Proposed changes
 
The changes have been implemented within the solve member of the class. During the pre-processing, we setup two new vectors vec_R_ and vec_Y_. The vector vec_R_ is the intial residual, whereas vec_Y_ contains our update (it is set to zero before the start of the FGMRES loop). These new vectors are essential replacing the rhs and the direct solution x from previous implementations. The main loop is largely the same, except we replace all references to x with vec_Y_ instead. After the loop we add the update to the initial guess, making sure to overwrite the update. The residual from this quantity is used to overwrite vec_R_. The norm of the new vec_R_ is computed and compared with the initial residual norm (saved during the pre-processing) and then the decision to overwrite x is made. 
 
 ## Checklist
 
 merging your code._
 
- [ x] All tests pass. Code tested on
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 
 ## Further comments
 
 _If this is a relatively large or complex change, kick off the discussion by explaining
 why you chose the solution you did and what alternatives you considered, etc._


shakedregev (changes requested):
Rebase and follow ReSolve guidelines.

examples/ExampleHelper.hpp
```diff
@@ -301,6 +312,156 @@ namespace ReSolve
...
+
+        // 4. Populate the vector with random numbers.
+        for (int i = 0; i < size; ++i)
+        {
+          // Use the 'at' method to safely access and modify the vector elements.
+          h_data[i] = dist(engine);
+        }
+
+        // 5. Calculate the Euclidean norm (L2 norm).
+        // The formula is: $||x||_2 = \sqrt{\sum_{i=1}^{n} |x_i|^2}$
```
shakedregev, line 342:
```suggestion
        // The formula is: $||x||_2 = \sqrt{\sum_{i=1}^{n} |x_i|^2}$.
```

Issue #295, opened 2025-06-02
Strictly enforce memory requirements.


shakedregev:
### Summary

There are some places in the code where memory is used in a sloppy manner. [This](https://github.com/ORNL/ReSolve/pull/293/files) is an example which likely occurs many times in the code.
For example, a vector may be allocated twice in certain cases or the code can try to delete something that was never allocated. We should enforce strict usage to make sure the code runs efficiently and that debugging is easy.

### Rationale

Sloppy memory management can make writing code easier, but it can create nasty bugs that are difficult to find. For example, one might expect data to be in one location, but it isn't because the function that was supposed to move it there quietly did nothing. In this case the incorrect result can be observed far downstream from the error and it can be very hard to track down.

### Description

A solution can build upon [this](https://github.com/ORNL/ReSolve/pull/293/files) and add assert statements. An alternative is to use [a check error wrapper](https://github.com/ORNL/ReSolve/blob/shaked/cpu_only_gmres_fix/resolve/cuda/cuda_check_errors.hpp) that pinpoints issues to a specific line.

### Additional information

@superwhiskers - do you have bandwidth for this?

pelesh:
CC @kswirydo

pelesh:
I think we need to decide whether to use and support sloppy allocations or to outright prohibit them. In the latter case, we would need to do more work far beyond this PR.

@kswirydo was a proponent of sloppy allocations and I see the value of it. Sloppy allocations can make user code be clear and compact. However, supporting and maintaining sloppy allocations requires significant developer overhead, which I am not sure we can afford.

@shakedregev makes a good point that debugging becomes more difficult with a sloppy memory management.

@PhilipFackler, @superwhiskers: do you care to chime in?

shakedregev:
Sloppy allocations improve compactness and are more forgiving, but they make for less clear code IMHO. It means that some functions do things that are not within their name or description.

superwhiskers:
we should explicitly not support them and remove existing instances of this from the codebase.

they may make user-facing code more compact, but they bloat implementation code and make user code less readable, in my opinion. if we were writing a library for a scripting language, i can see why we may want this, but because we're a c++ library, i think the more predictable behavior is that the library does not implicitly do things for me that i forgot to do.

we can address the issue of making code easier to write by having well-written error messages (emitted through debug assertions or plain errors) where we would've implicitly allocated, or we can redesign parts of the library so that it is impossible for code written that way to compile to begin with.

Issue #19, opened 2023-10-18
Should memory allocation check if the pointer is NULL?


pelesh:
At several places in sparse matrix classes we check if the pointer is `nullptr` before we allocate memory and assign it to that pointer. Here is one such example:
```c++
        if (d_col_data_ == nullptr) {
          mem_.allocateArrayOnDevice(&d_col_data_, n_ + 1); 
        }
        if (d_row_data_ == nullptr) {
          mem_.allocateArrayOnDevice(&d_row_data_, nnz_current); 
        }
        if (d_val_data_ == nullptr) {
          mem_.allocateArrayOnDevice(&d_val_data_, nnz_current); 
        }
```
Perhaps we should consider throwing exception when the pointer is not `nullptr`? This way potential bugs may remain undetected, because if the pointer is not a null, this does not necessarily mean that the allocated memory is of the size we need. In such case we may end up with undefined behavior such as writing past the end of array. More defensive programming would be to send an error message and throw an exception in situations like this.

CC @rothpc @maksud

PR #383, opened 2025-10-07
Disallow sloppy calls to functions that allocate memory when they shouldn't.

shakedregev:
## Description
 
 _Memory allocation can be incidental in ReSolve rather than explicitly defined by the user. Multiple calls to functions that should only be called once, or calls to function without first calling a prerequisite function are allowed. Addresses (no commitment that it fully closes) #295 and #19. I'd like your opinion on whether you think it closes these issues. Currently the code allows the user to set a vector to a constant without allocating it. I am ambivalent, but I do think it should be allowed over all. Let me know your thoughts._

 ## Proposed changes
 
 _Fixed typos, removed the ability to sloppy synch or sloppily call functions out of order._
 
 ## Checklist
 
 merging your code._
 
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
- [x] No merge conflicts.
- [x] The changes are included within the changelog.

### DIFF
--- a/examples/ExampleHelper.hpp
+++ b/examples/ExampleHelper.hpp
@@ -142,14 +142,10 @@ namespace ReSolve
                        ReSolve::vector::Vector* r,
                        ReSolve::vector::Vector* x)
       {
+        assert(res_ != nullptr && "resetSystem should be called after setSystem");
         A_ = A;
         r_ = r;
         x_ = x;
-        if (res_ == nullptr)
-        {
-          res_ = new ReSolve::vector::Vector(A->getNumRows());
-        }
-
         computeNorms();
       }
 
--- a/resolve/LinSolverDirectCuSolverRf.cpp
+++ b/resolve/LinSolverDirectCuSolverRf.cpp
@@ -79,6 +79,7 @@ namespace ReSolve
     if (setup_completed_)
     {
       out::error() << "LinSolverDirectCuSolverRf::setup should only be called one." << std::endl;
+      return 1;
     }
 
     if (d_P_ == nullptr)
@@ -88,6 +89,7 @@ namespace ReSolve
     else
     {
       out::error() << "d_P_ should be nullptr on call to LinSolverDirectCuSolverRf::setup." << std::endl;
+      return 1;
     }
 
     if (d_Q_ == nullptr)
@@ -97,6 +99,7 @@ namespace ReSolve
     else
     {
       out::error() << "d_Q_ should be nullptr on call to LinSolverDirectCuSolverRf::setup." << std::endl;
+      return 1;
     }
 
     if (d_T_ == nullptr)
@@ -106,6 +109,7 @@ namespace ReSolve
     else
     {
       out::error() << "d_T_ should be nullptr on call to LinSolverDirectCuSolverRf::setup" << std::endl;
+      return 1;
     }
 
     mem_.copyArrayHostToDevice(d_P_, P, n);
--- a/resolve/LinSolverDirectCuSolverRf.cpp
+++ b/resolve/LinSolverDirectCuSolverRf.cpp
@@ -78,7 +78,7 @@ namespace ReSolve
 
     if (setup_completed_)
     {
-      out::error() << "LinSolverDirectCuSolverRf::setup should only be called one." << std::endl;
+      out::error() << "Trying to setup LinSolverDirectCuSolverRf, but the setup has been already done! " << std::endl;
       return 1;
     }
 
@@ -88,7 +88,7 @@ namespace ReSolve
     }
     else
     {
-      out::error() << "d_P_ should be nullptr on call to LinSolverDirectCuSolverRf::setup." << std::endl;
+      out::error() << "Trying to allocate permutation vector P in " << __func__ << " in LinSolverDirectCuSolverRf, but the permutation vector P is already allocated! " << std::endl;
       return 1;
     }
 
@@ -98,7 +98,7 @@ namespace ReSolve
     }
     else
     {
-      out::error() << "d_Q_ should be nullptr on call to LinSolverDirectCuSolverRf::setup." << std::endl;
+      out::error() << "Trying to allocate permutation vector Q in " << __func__ << " in LinSolverDirectCuSolverRf, but the permutation vector Q is already allocated! " << std::endl;
       return 1;
     }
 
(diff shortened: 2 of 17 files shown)


resolve/SystemSolver.cpp
```diff
@@ -393,32 +389,43 @@ namespace ReSolve
...
     }
-
     if (refactorizationMethod_ == "glu" || refactorizationMethod_ == "cusolverrf" || refactorizationMethod_ == "rocsolverrf")
     {
       is_solve_on_device_ = true;
       return refactorizationSolver_->refactorize();
     }
-
+    out::error() << "Refactorization method not set to a valid method!" << std::endl;
     return 1;
```
pelesh:
```suggestion
    out::error() << "Refactorization method " << refactorizationSolver_ << " not recognized!\n"
                 << "Available options: klu, glu, cusolverrf, rocsolverrf.\n";
    return 1;
```

resolve/SystemSolver.cpp
```diff
@@ -393,32 +389,43 @@ namespace ReSolve
...
+      out::error() << "System matrix not set!\n";
+      return 1;
+    }
     if (factorizationMethod_ == "klu")
     {
       is_solve_on_device_ = false;
       return factorizationSolver_->factorize();
     }
+    out::error() << "Factorization method not set to a valid method!" << std::endl;
     return 1;
```
pelesh:
```suggestion
    out::error() << "Factorization method " << factorizationSolver_ << " not recognized!\n"
                 << "Available options: klu.\n";
    return 1;
```

resolve/SystemSolver.cpp
```diff
@@ -393,32 +389,43 @@ namespace ReSolve
       factorizationSolver_->setup(A_);
       return factorizationSolver_->analyze();
     }
+    out::error() << "Analyze method not set to a valid method!" << std::endl;
     return 1;
```
pelesh:
```suggestion
    out::error() << "Analysis method " << factorizationSolver_ << " not recognized!\n"
                 << "Available options: klu.\n";
return 1;
```

resolve/LinSolverDirectRocSolverRf.cpp
```diff
@@ -82,11 +82,21 @@ namespace ReSolve
     {
       mem_.allocateArrayOnDevice(&d_P_, n);
     }
+    else
+    {
+      out::error() << "d_P_ should be nullptr on call to LinSolverDirectRocSolverRf::setup." << std::endl;
+      return 1;
```
pelesh:
```suggestion
      out::error() << "Trying to allocate permutation vector P in " << __func__ << " but the vector is already allocated!\n";
      return 1;
```

resolve/LinSolverDirectRocSolverRf.cpp
```diff
@@ -82,11 +82,21 @@ namespace ReSolve
...
+    }
 
     if (d_Q_ == nullptr)
     {
       mem_.allocateArrayOnDevice(&d_Q_, n);
     }
+    else
+    {
+      out::error() << "d_Q_ should be nullptr on call to LinSolverDirectRocSolverRf::setup." << std::endl;
+      return 1;
```
pelesh:
```suggestion
      out::error() << "Trying to allocate permutation vector Q in " << __func__ << " but the vector is already allocated!\n";
      return 1;
```

resolve/vector/Vector.cpp
```diff
@@ -668,33 +668,33 @@ namespace ReSolve
...
         cpu_updated_[j] = true;
         gpu_updated_[j] = false;
         break;
       case DEVICE:
         if (d_data_ == nullptr)
         {
-          mem_.allocateArrayOnDevice(&d_data_, n_capacity_ * k_);
-          owns_gpu_data_ = true;
+          out::error() << "Trying to set vector device values, but the values are not allocated!" << std::endl;
+          return 1;
```
pelesh:
```suggestion
          out::error() << "Trying to set vector values on the device, but the memory is not allocated!" << std::endl;
          return 1;
```

resolve/matrix/io.cpp
```diff
@@ -386,7 +386,11 @@ namespace ReSolve
         Logger::error() << "Empty input to updateArrayFromFile function ..." << std::endl;
         return;
       }
-
+      if (p_rhs == nullptr)
+      {
+        Logger::error() << "Null pointer to array in updateArrayFromFile function ..." << std::endl;
+        return;
```
pelesh:
```suggestion
        Logger::error() << "Memory not allocated for updateArrayFromFile function ...\n";
        return;
```

examples/ExampleHelper.hpp
```diff
@@ -142,14 +142,10 @@ namespace ReSolve
                        ReSolve::vector::Vector* r,
                        ReSolve::vector::Vector* x)
       {
+        assert(res_ != nullptr && "resetSystem should be called after setSystem");
         A_ = A;
```
pelesh, line 146: I would not touch this file for now. It needs more refactoring than this.

resolve/LinSolverDirectCuSolverRf.cpp
```diff
@@ -75,31 +76,41 @@ namespace ReSolve
...
 
-    // Remember - P and Q are generally CPU variables!
-    // Factorization data is stored in the handle.
-    // If function is called again, destroy the old handle to get rid of old data.
     if (setup_completed_)
     {
-      cusolverRfDestroy(handle_cusolverrf_);
-      cusolverRfCreate(&handle_cusolverrf_);
+      out::error() << "LinSolverDirectCuSolverRf::setup should only be called one." << std::endl;
+      return 1;
```
pelesh:
```suggestion
      out::error() << "Trying to setup LinSolverDirectCuSolverRf, but the setup has been already done!\n";
      return 1;
```

pelesh: Similar suggestions to error messages for other allocations below.

pelesh (commented):
Some suggestions on error messaging. In principle, I would tell users what (generic) procedure has faild/was called in error rather than specifying a cryptic variable name. We could use `__line__` and `__func__` to provide a location of the error in the code.

examples/ExampleHelper.hpp
shakedregev: The change is very simple and is better than what we have currently. We can debate the error message, but the user should not be allowed to reset the system without first setting it.

shakedregev:
Thanks for the suggestions. I will implement them locally because they touch more than the comments (make sure there are no typos in variable names for example) and push them. The one exception is the suggestion I disagreed with explicitly.

pelesh:
More comprehensive solution merged in #412 and #416.

PR #392, opened 2025-11-07
cmake-format

nkoukpaizan:
## Description

This adds CMake formatting.  Adapted from [ExaGO](https://github.com/pnnl/ExaGO/blob/develop/.cmake-format.py)

 ## Proposed changes

- Added `.cmake-format.py` defining the desired format
- Added a stage in `.pre-commit-config.yaml` to apply the format
 
 ## Checklist
 
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
- [x] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra` (apart from known issues).
- [N/A] The new code follows Re::Solve style guidelines.
- [N/A] The new code is documented.
 
 ## Further comments
Please review the formatting that was applied. We can try to change whatever doesn't look the way we want.

### DIFF
--- a/cmake/ReSolveFindHipLibraries.cmake
+++ b/cmake/ReSolveFindHipLibraries.cmake
@@ -1,6 +1,5 @@
 # Exports target `resolve_hip` which finds all hip libraries needed by resolve.
 
-
 add_library(resolve_hip INTERFACE)
 
 find_package(hip REQUIRED)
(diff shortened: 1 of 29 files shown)


PhilipFackler (approved):
Yes! Love this! I will refrain from being opinionated about the formatting. Other folks should weigh in with that. But I'll just be happy for it to be consistent. :)

pelesh:
This is great! Can the CMake format be configured so that when we have multiple arguments in a function, each argument is on a separate line. For example this
```cmake
target_link_libraries(
  resolve_cuda INTERFACE CUDA::cusolver 
                         CUDA::cublas
                         CUDA::cusparse
                         CUDA::cudart
)
```
is more clear code than this:
```cmake
target_link_libraries(
  resolve_cuda INTERFACE CUDA::cusolver CUDA::cublas CUDA::cusparse
                         CUDA::cudart
)
```
I am not sure how easy or difficult is to do this and I agree with @PhilipFackler that consistency is the primary goal here.

Feel free to merge as is or adjust formatting as you see fit.

pelesh (approved):

cmake/ReSolveFindHipLibraries.cmake
```diff
@@ -1,36 +1,33 @@
...
+  target_link_libraries(
+    resolve_hip INTERFACE rocprofiler-sdk::rocprofiler-sdk
+                          rocprofiler-sdk-roctx::rocprofiler-sdk-roctx
   )
 endif()
 
-# HIP/ROCm targets still don't have include directories set correctly
-# We need this little hack for now :/
+# HIP/ROCm targets still don't have include directories set correctly We need
+# this little hack for now :/
```
shakedregev:
```suggestion
# HIP/ROCm targets still don't have include directories set correctly. We need
# this little hack for now :/
```

shakedregev (approved):
Looks good. I ran independent tests and they pass on both platforms.

PR #405, opened 2026-01-30
Add preconditioner interface

### LINKED ISSUE #114: Create a preconditioner interface class.
I think it would be a good idea to create abstract interface for iterative solver preconditioners and expose them to users. Current preconditioners we use should inherit from that interface, as well. I suggest something like this:
- The preconditioner class implements method(s) `applyPreconditioner()`, possibly overloaded.
- Preconditioner interface is exposed to users so they can write their own preconditioners.
- Iterative solver base class should implement method `setPreconditioner`, which should basically set a pointer to external preconditioner instance.
- We could wrap preconditioners we currently use in this abstract preconditioner class.

CC @kswirydo

kakeueda:
## Description
 
This adds a preconditioner base class and refactors `SystemSolver` to use it, removing the current coupling to `LinSolverDirect` and enabling custom preconditioner implementations.

Closes #114 
 
cc @shakedregev 

 ## Proposed changes

Added a `Preconditioner` base class with the following methods:
  - `setup()`  
  - `apply()` (calls the preconditioner)
  - `reset()` (for updated matrix values with the same sparsity pattern)
 
 ## Checklist
 
 
 

     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 ## Further comments


shakedregev:
@KakeruUeda - it seems some tests are not passing. I will wait before I review this.

pelesh (approved):
Clean implementation consistent with the rest of the code. All tests pass.

PR #408, opened 2026-02-23
ABBA GMRES Solver

JeffZ594:
## Description
 
 

Adds ABBA GMRES functionality to the FGMRES class 

 
 
  
 
@pelesh 

 ## Proposed changes
 
 
 
ABBA GMRES (Left and Right Preconditioned) functionality is added which allows users to precondition from either the left or the right using user provided preconditioning matrices.

 ## Checklist
 
 
 

     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
- [ ] I have manually run the non-experimental examples and verified that residuals are close to machine precision. (In your build directory run:
     - [ ] CPU backend
     - [ ] CUDA backend
     - [ ] HIP backend
 
 ## Further comments
 
I do not use the Preconditioner class as the FGMRES class itself has to have access to the preconditioner matrix pointer as the solve function has to be modified for the different initial residual used by BA GMRES.

I also believe that the code would become unnecessarily complicated if I were to use the Preconditioner class as knowledge of whether AB or BA GMRES is used must be known to the FGMRES class itself and passing this information to another class seems unnecessary given the amount of modification already needed in the class itself and the somewhat minimal changes needed to use matvec product for preconditioning.


pelesh (commented):
Very nice work, thank you for your contribution!

A couple of preliminary comments:
- We need to use preconditioner interface to have uniform access to all current and future preconditioner. If the the existing preconditioner interface is inadequate, please suggest how to modify it.
- I am not sure if we need to include `trips_code` directory in Re::Solve repository. Test matrices need to be stored with other Re::Solve test objects. Other information could be supplied in a README file with links to trips repository.
- Test cases need to be documented better, including properties of the system matrix and preconditioners. This could go in the same README file.

Once these are addressed we can set this PR as ready for review.

CC @mpasha3 @shakedregev

JeffZ594:
Understood, I will organize the files properly and think about how to use the preconditioner class.

I think the best way to approach this would be to have the initial residual calculated by the preconditioner class itself. Do you think this would be a satisfactory way to modify the preconditioner class?

kakeueda:
I think we could add a PreconditionerMatvec class for user-defined preconditioners which wraps the current matvec operation in an apply() method. When computing the initial residual, we can call preconditioner->getSide() and preconditioner->getMatrix() instead.

It may also be useful to add some AB/BA operations to the preconditioner base class. This petsc's implementation ([PCApplyBAorAB](https://petsc.org/release/src/ksp/pc/interface/precon.c.html#PCApplyBAorAB)) might help.

JeffZ594:
Thank you for the suggestion! I will look into the petsc implementation.

JeffZ594:
I have implemented ABBA GMRES using the preconditioner interface. 

@pelesh for the test case I have, should I include the script I have to run the test and the python code to plot the solutions in the pull request? If so should they be put in the same folder as the other Re::Solve test objects?

kakeueda:
@pelesh I accidentally  re-requested review. Please ignore this.

resolve/PreconditionerMatvec.cpp
```diff
@@ -0,0 +1,121 @@
...
+    }
+    // BA GMRES Preconditioner
+    else
+    {
+      matrix_handler_->matvec(A_, rhs, x, &ONE, &ZERO, memspace_);
+      return 0;
+    }
+    out::error() << "Appication of Preconditioner Failed\n";
+    return 1;
+  }
```
kakeueda:
```suggestion
  int PreconditionerMatvec::apply(vector_type* rhs, vector_type* x)
  {
    using namespace constants;
    matrix_handler_->matvec(B_, rhs, x, &ONE, &ZERO, memspace_);
    return 0;
  }
```
It's better to write it this way because else{...} isn't actually applying the preconditioner. Instead, you can swap the order of the precV and matvec in the FGMRES solve() function

resolve/LinSolverIterativeFGMRES.cpp
```diff
@@ -200,11 +210,20 @@ namespace ReSolve
...
+        // Right preconditioned
+        if (preconditioner_->getSide() == "right")
+        {
+          matrix_handler_->matvec(A_, &vec_z, &vec_v, &ONE, &ZERO, memspace_);
+        }
+        // Left Preconditioned
+        else
+        {
+          matrix_handler_->matvec(preconditioner_->getPrec(), &vec_z, &vec_v, &ONE, &ZERO, memspace_);
+        }
```
kakeueda, line 232:
```suggestion
        if (preconditioner_->getSide() == "right")
        {
          this->precV(&vec_v, &vec_z);
          mem_.deviceSynchronize();
          vec_v.setData(vec_V_->getData(i + 1, memspace_), memspace_);
          matrix_handler_->matvec(A_, &vec_z, &vec_v, &ONE, &ZERO, memspace_);
        }
        // Left Preconditioned
        else
        {
          matrix_handler_->matvec(A_, &vec_z, &vec_v, &ONE, &ZERO, memspace_);
          mem_.deviceSynchronize();
          vec_v.setData(vec_V_->getData(i + 1, memspace_), memspace_);
          this->precV(&vec_v, &vec_z);
        }
```

Swapping the matvec() and precV() should work with the updated apply(). This is also what petcs does.

resolve/Preconditioner.hpp
```diff
@@ -32,8 +35,13 @@ namespace ReSolve
...
     virtual ~Preconditioner();
 
-    virtual int setup(matrix_type* A)                   = 0;
-    virtual int apply(vector_type* rhs, vector_type* x) = 0;
-    virtual int reset(matrix_type* /* A */);
+    virtual int          setup(matrix_type* A)                   = 0;
+    virtual int          apply(vector_type* rhs, vector_type* x) = 0;
+    virtual int          reset(matrix_type* /* A */);
+    virtual std::string  getSide(); // Gets the preconditioning side
+    virtual matrix_type* getPrec(); // Dummy function only implemented by PreconditionerMatvec
```
kakeueda:
```suggestion
    virtual matrix_type* getPrecMatrix(); // Get the preconditioner matrix
```

I would write it this way because it returns the matrix and other preconditioners might need it.

resolve/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ set(ReSolve_SRC
     SystemSolver.cpp
     Preconditioner.cpp
     PreconditionerLU.cpp
+    PreconditionerMatvec.cpp
```
kakeueda: This is my fault, but it’s probably better to rename this file (and class) to reflect its purpose.

kakeueda:
Good improvement overall. I left a few suggestions.

tests/functionality/testTripsProblem.cpp
```diff
@@ -0,0 +1,214 @@
...
+  status = FGMRES.solve(vec_rhs, &vec_x);
+  error_sum += status;
+
+  // Print result summary and check solution
+  std::cout << "\n  BAGMRES results: \n"
+            << "\t Hardware backend:                              : "
+            << hwbackend << "\n";
+  helper.printIterativeSolverSummary(&FGMRES);
+  std::ofstream BA_output_file("BA_output.mtx");
+  error_sum += ReSolve::io::writeVectorToFile(&vec_x, BA_output_file);
```
pelesh, line 205: Why do you output vector here? Is this file used anywhere?

tests/functionality/testTripsProblem.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+  isTestPass(error_sum, "Test ABBAGMRES on " + hwbackend + " device");
```
pelesh, line 207: You would want to test the solver for consistency here, similar to how it is done in [`randGmres`](https://github.com/ORNL/ReSolve/blob/develop/tests/functionality/testRandGmres.cpp#L175) example. This will give a proper pass/fail criterion.

pelesh (changes requested):
Looks good. Two straightforward changes are needed before we can merge this:
- Add check for consistency of the solution to `testTripsProblem`. Make sure ||Ax-b|| < machine eps.
- Please remove `trips_code` directory. Instead you may want to add readme file with links to Trips library and results obtained with Re::Solve.

tests/functionality/testTripsProblem.cpp
JeffZ594: The file is used to plot the solution obtained against the true solution. This is more of an extra visualization to see how the solver performs.

tests/functionality/testTripsProblem.cpp
JeffZ594: The way the trips test problem is set up I believe it is not possible to obtain an error close to machine epsilon due to the noise being added. This is part of the reason why I use the plots because for iterations above 6 the solution starts to diverge even though the error is lower.

Originally this test was meant to be used to demonstrate the effect of regularization. Would it be more appropriate to use the same test problem as randGmres to show the consistency of the solver? This way the properties of the matrix are also known.

pelesh:
@JeffZ594: Also, please add `testTripsProblem` as another functionality test in CMake. Make sure test files are accessible to the test. See implementation of [`testRefactor`](https://github.com/ORNL/ReSolve/blob/develop/tests/functionality/testRefactor.cpp), for example.

pelesh:
@JeffZ594, why do you input both $A$ and $A^T$? You could use transpose function from Re::Solve to get $A^T$.

JeffZ594:
@pelesh The Re::Solve transpose function is used to get A^T. I give 2 A files to allocate memory for 2 matrices but they are both the same A matrix. 

Also, to add on to the comment about the test problem in testTripsProblem, I could also set the test pass tolerance to the error that I expect after 6 iterations, but I think making a different test problem may be more appropriate since the ill-posed nature of the Trips problem is not really relevant to the ABBA solver.

Issue #85, opened 2023-11-28
Setting performance tuning parameters for GPU kernels


pelesh:
Perhaps we should consider how to replace magic numbers used for performance parameters in GPU kernels with something that can be more easily adjusted, if needed. It would be good to allow expert user to tune code performance by setting these parameters.

_Originally posted by @pelesh in https://github.com/ORNL/ReSolve/pull/83#discussion_r1408097301_

Issue #18, opened 2023-10-18
Naming scheme for vector classes


pelesh:
It might be a good idea to decide on vector naming schemes. I think it would be important to have these schemes analog to what we settled on for matrices. Here are a few suggestions:
- Rename namespace `vector` to `multivector`. Then rename `Vector` class to `Dense` in analogy with matrix objects. The class will be called as `multivector::Dense`, i.e. this convention would include that `using namespace multivector` is never invoked.
- Alternatively, we could keep the namespace as `vector` and rename the class `MultiDense`, so the invocation of the type name would be `vector::MultiDense`. This name is a little bit cumbersome, but it also allows us to derive single vector from multivector class trivially and then have another type name `vector::SingleDense`. This may provide more flexibility.

Once we decide, it is straightforward to update names of classes and files.

@kswirydo @cameronrutherford @rothpc  please chime in.

pelesh:
Looking at `vec_dev` branch and the work @kswirydo charted there, I suggest following vector classes naming scheme:
```c++
namespace vector { namespace dense {
  class Multi {
    // dense multivector class declaration
  };
}}
```
The vector object should be called with both namespaces preceding it explicitly:
```c++
vector::dense::Multi vec;
```
This way, namespaces protect from name conflicts for vector classes and help code readability.

Alternatively, with this naming scheme we could take advantage of `using` statement like this:
```c++
using vector_type = vector::dense::Multi;
// some code
vector_type vec;
```
Here we define type `vector_type` that we want to use in this scope exclusively. We have a line that shows what `vector_type` is, and we use it as a short hand notation through the rest of the scope. If we change the first line, we are guaranteed to have a wholesale change within the entire scope.

On the other hand, `using` statements like this should be **discouraged** with this notation:
```c++
using namespace vector::dense;
// 300 lines of code
Multi vec;
```
If we change using statement here, we still need to verify that each instance of `Multi` (1) exists in the new namespace and (2) is used correctly.

A single vector class can be derived from `Multi` vector almost trivially:
```c++
namespace vector { namespace dense {
  class Single : public Multi {
    public:
      Single() : Multi(numVectors = 1)
      {}
    // some code
  };
}}
```

If we decide to implement e.g. our own matrix factorization, we will need a sparse vector object. In that case we could extend this naming scheme and define sparse vector class in different namespace like this:
```c++
namespace vector { namespace sparse {
  class Multi {
    // sparse multivector class declaration
  };
}}
```

[...]

PR #411, opened 2026-02-25
Use BLAS naming for VectorHandler methods

### LINKED ISSUE #410: Consider reviewing naming scheme for `VectorHandler` methods
It would be good to consider if methods in `VectorHandler` class could have more intuitive names:
I just copied from the documentation which is incorrect. We need to write what the functions actually do.
- `massDot2Vec` computes $r = V^T x$, where $V$ is $n \times k$ and $x$ is a $k \times 2$ multivector (dense matrix).
- `massAxpy` computes $y := y + \alpha x$, where $x$ is $n \times k$ and $\alpha$ is $k \times 1$ (multi)vector.
- `scal` scales vector by a scalar value.
- `scale` is elementwise vector product used to implement scaling a vector by a diagonal matrix.

The remaining functions names `axpy`, `dot` and `gemv` should be recognizable to anyone with some knowledge of BLAS.

@maksud @rothpc @shakedregev @nkoukpaizan please chime in with your suggestions.

pelesh:
## Description
 
 Updates names of methods in `VecorHandler` to match BLAS naming convention.
 
Closes #410

 CC @alexander-novo

 ## Proposed changes
 
- Names of vector operations changed to match BLAS naming scheme as close as possible.
- Updated and corrected Doxygen docs.
- Scalar arguments passed by value.
- `assert` statements added to aid debugging.

 ## Checklist
 
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 ## Further comments
 
 Assert statements in `gemv` fail for some reason. Need to investigate that more; out of scope here.


alexander-novo:
Shouldn't the infinity norm be `amax`? `iamax` returns the index of the maximum value.

pelesh:
> Shouldn't the infinity norm be `amax`? `iamax` returns the index of the maximum value.

Good catch. Addressed in https://github.com/ORNL/ReSolve/pull/411/commits/8310c9f656d006c849b14a71c318267ee347384e.

resolve/cuda/cudaKernels.cu
```diff
@@ -308,7 +308,7 @@ namespace ReSolve
      */
     void mass_axpy(index_type n, index_type i, const real_type* x, real_type* y, const real_type* alpha)
```
kakeueda: Maybe rename it to axpy_multi (or axpyMulti)? Any reason for using the underscore here?

resolve/vector/VectorHandlerCuda.cpp
```diff
@@ -73,8 +80,9 @@ namespace ReSolve
     cublasStatus_t st            = cublasDscal(handle_cublas, x->getSize(), alpha, x->getData(memory::DEVICE), 1);
```
kakeueda: Looks like you forgot to change this part

kakeueda (commented):
All tests passed with the CUDA backend after fixing VectorHandlerCuda::scal

kakeueda (changes requested):

resolve/cuda/cudaKernels.cu
pelesh: I didn't touch GPU kernels in this PR because they need more work but harmonizing names with calling functions is a good idea. Thanks!

resolve/cuda/cudaKernels.cu
```diff
@@ -308,7 +308,7 @@ namespace ReSolve
...
     {
-      kernels::massAxpy3<<<(n + 384 - 1) / 384, 384>>>(n, i, x, y, alpha);
+      kernels::axpyMulti3<<<(n + 384 - 1) / 384, 384>>>(n, i, x, y, alpha);
```
shakedregev, line 311: Is the reason for this number 3*128 and it's just good for the hardware?

resolve/vector/VectorHandler.cpp
```diff
@@ -201,14 +212,22 @@ namespace ReSolve
...
                            memory::MemorySpace memspace)
   {
     using namespace ReSolve::memory;
+
+    // TODO: Remove n as the argument becuase it must always be n = V->getSize()
+    assert(n == V->getSize() && "gemv: n does not match the number of rows in V");
+
+    // TODO: These assertions should hold but they do not. Investigate more.
+    // assert((transpose == 'T' && V->getSize() == y->getSize()) && "gemv: size mismatch! size of V^T does not match size of y");
+    // assert(((transpose == 'N') && (V->getSize() == x->getSize())) && "gemv: size mismatch! size of V does not match size of x");
```
shakedregev: Why not fix these before merging?

resolve/vector/VectorHandler.cpp
```diff
@@ -222,56 +241,75 @@ namespace ReSolve
...
-   * @param[in] V        - Multivector; k vectors size n x 1 each
-   * @param[in] k        - Number of vectors in V
-   * @param[in] x        - Multivector; 2 vectors size n x 1 each
-   * @param[out] res     - Multivector; 2 vectors size k x 1 each (result is returned in res)
-   * @param[in] memspace - String containg memspace (cpu or cuda or hip)
+   * @param[in] V        - Multivector; k vectors of size n x 1 each
+   * @param[in] k        - Number of vectors in V to use
+   * @param[in] x        - Multivector; 2 vectors of size n x 1 each
+   * @param[out] res     - Multivector; 2 vectors size k x 1 each
+   * @param[in] memspace - String containg memspace (cpu, cuda or hip)
```
shakedregev:
```suggestion
   * @param[in] memspace - String containg memspace (cpu, cuda, or hip)
```

resolve/vector/VectorHandler.cpp
```diff
@@ -222,56 +241,75 @@ namespace ReSolve
...
+   * @pre _size_ > 0, _k_ > 0, size = x->getSize().
+   * @pre _res_ needs to be allocated to k x 2 size.
    *
    */
-  void VectorHandler::massDot2Vec(index_type size, vector::Vector* V, index_type k, vector::Vector* x, vector::Vector* res, memory::MemorySpace memspace)
+  void VectorHandler::dot2Multi(index_type          size,
+                                vector::Vector*     V,
+                                index_type          k,
+                                vector::Vector*     x,
+                                vector::Vector*     res,
```
shakedregev, line 296: Should follow Householder notation (V should be v)

resolve/vector/VectorHandlerCpu.cpp
```diff
@@ -205,11 +210,11 @@ namespace ReSolve
    * @pre   _k_ > 0, _size_ > 0, _size_ = x->getSize()
    *
    */
-  void VectorHandlerCpu::massAxpy(index_type      size,
-                                  vector::Vector* alpha,
-                                  index_type      k,
-                                  vector::Vector* x,
-                                  vector::Vector* y)
+  void VectorHandlerCpu::axpyMulti(index_type      size,
+                                   vector::Vector* alpha,
```
shakedregev, line 226: Greek letters should be scalars

shakedregev (changes requested):
Fix the failing asserts and name convention violations.

resolve/vector/VectorHandler.cpp
pelesh: Some of the "vectors" are in fact multivectos. Capital `V` indicates that the instance contains multiple vectors. We should probably decide on notation and even how we call the vector class (`vector::Vector` or `vector::Multivector`). This is tracked in #18.

resolve/vector/VectorHandlerCpu.cpp
pelesh: In this case using greek letter is to signify that these are "non-interacting" scalars that are stored in a vector object. This kernel is doing embarrassingly parallel vector `axpy`s. It is probably a good idea if we decide these conventions in #18, together with vector naming scheme.

resolve/cuda/cudaKernels.cu
pelesh: This is a known issue tracked in #85. It is out of scope of this PR.

resolve/vector/VectorHandlerCuda.cpp
pelesh: Fixed in https://github.com/ORNL/ReSolve/pull/411/commits/455c25ab1e3f6b5806ff6430cbbd953d47711aa1

kakeueda (approved):

resolve/cuda/cudaKernels.cu
pelesh: Addressed in https://github.com/ORNL/ReSolve/pull/411/commits/5dbd95c85bfe3d42edba6cdcc77b76e1474ea9a0

resolve/vector/VectorHandler.cpp
pelesh: It is a bit out of scope of this PR. The issue encountered is tracked in #415.

pelesh: Addressed in https://github.com/ORNL/ReSolve/pull/411/commits/10100828fd344547f91d812c78cbb634f24264b9. The `VectorHandler::gemv`method performs correctly, but the way it is used in `GramSchmidt` class is sloppy. We need to refactor `GramSchmidt` before we can add more rigorous protections to `gemv`.

resolve/cuda/cudaKernels.cu
```diff
@@ -287,12 +287,12 @@ namespace ReSolve
      * value of Tv5?
      * @todo Should we use dynamic shared memory here instead?
      */
-    void mass_inner_product_two_vectors(index_type       n,
-                                        index_type       i,
-                                        const real_type* vec1,
-                                        const real_type* vec2,
-                                        const real_type* mvec,
-                                        real_type*       result)
+    void dot_2_multi(index_type       n,
```
shakedregev, line 290: Why does this one use underscores instead of camel case.

resolve/cuda/cudaKernels.cu
```diff
@@ -306,9 +306,9 @@ namespace ReSolve
      * @param[out] y     - (n x (i+1)) multivector
      * @param[in]  alpha - ((i+1) x 1) vector
      */
-    void mass_axpy(index_type n, index_type i, const real_type* x, real_type* y, const real_type* alpha)
+    void axpy_multi(index_type n, index_type i, const real_type* x, real_type* y, const real_type* alpha)
```
shakedregev, line 309: So the wrapper uses underscores and the kernel uses camel?

resolve/vector/VectorHandler.cpp
```diff
@@ -182,33 +185,50 @@ namespace ReSolve
...
+                           const real_type     beta,
                            vector::Vector*     V,
                            vector::Vector*     y,
                            vector::Vector*     x,
                            memory::MemorySpace memspace)
   {
     using namespace ReSolve::memory;
+
+    // TODO: Remove n as the argument becuase it must always be n = V->getSize()
+    assert(n == V->getSize() && "gemv: n does not match the number of rows in V");
```
shakedregev: Why not just set `n=V->getSize()` and not accept it as an argument?

resolve/vector/VectorHandler.cpp
```diff
@@ -222,56 +242,75 @@ namespace ReSolve
...
     }
     y->setDataUpdated(memspace);
   }
 
   /**
-   * @brief mass (bulk) dot product i.e,  V^T x, where V is n x k dense multivector (a dense multivector consisting of k vectors size n)
-   *        and x is k x 2 dense multivector (a multivector consisiting of two vectors size n each)
+   * @brief Multivector dot product, i.e  V^T x
+   *
+   * Computes V^T x with k vectors from multivector V. Result is storred
```
shakedregev:
```suggestion
   * Computes V^T x with k vectors from multivector V. Result is stored
```

resolve/vector/VectorHandler.cpp
```diff
@@ -222,56 +242,75 @@ namespace ReSolve
...
-   * @param[in] V        - Multivector; k vectors size n x 1 each
-   * @param[in] k        - Number of vectors in V
-   * @param[in] x        - Multivector; 2 vectors size n x 1 each
-   * @param[out] res     - Multivector; 2 vectors size k x 1 each (result is returned in res)
-   * @param[in] memspace - String containg memspace (cpu or cuda or hip)
+   * @param[in] V        - Multivector; k vectors of size n x 1 each
+   * @param[in] k        - Number of vectors in V to use
+   * @param[in] x        - Multivector; 2 vectors of size n x 1 each
+   * @param[out] res     - Multivector; 2 vectors size k x 1 each
+   * @param[in] memspace - String containg memspace (cpu, cuda, or hip)
```
shakedregev, line 286:
```suggestion
   * @param[in] memspace - String containing memspace (cpu, cuda, or hip)
```

shakedregev (commented):
These are important changes. There are a few naming things that slipped.

resolve/cuda/cudaKernels.cu
pelesh: My idea was to have GPU function names in C-style, both drivers and kernels. Lots of kernels were grandfathered in the code without changing their names. We need a separate issue and PR to address that.

resolve/vector/VectorHandler.cpp
pelesh: You are absolutely right. Done in https://github.com/ORNL/ReSolve/pull/411/commits/7deaa1631ff114562e2ae66beb5ff7e01bbdf5c5

resolve/vector/VectorHandler.cpp
pelesh: Fixed in https://github.com/ORNL/ReSolve/pull/411/commits/7deaa1631ff114562e2ae66beb5ff7e01bbdf5c5

resolve/cuda/cudaKernels.cu
pelesh: Need to address disregard for any naming conventions for GPU kernels and drivers in a separate issue/PR.

shakedregev (approved):
Approved after testing on Cuda and RocM

resolve/vector/VectorHandler.cpp
```diff
@@ -182,96 +185,128 @@ namespace ReSolve
...
   /**
-   * @brief mass (bulk) axpy i.e, y = y - x*alpha where  alpha is a vector
+   * @brief Multivector axpy: y: = y + \sum_i alpha_i x_i
    *
    * @param[in] size number of elements in y
    * @param[in] alpha vector size k x 1
-   * @param[in] x (multi)vector size size x k
+   * @param[in] x (multi)vector [size x k]
    * @param[in,out] y vector size size x 1 (this is where the result is stored)
    * @param[in] memspace string containg memspace (cpu or cuda or hip)
```
shakedregev:
```suggestion
   * @param[in] memspace string containg memspace (cpu, cuda, or hip)
```

PR #412, opened 2026-02-26
Add const getData methods to Vector class

pelesh:
## Description
 
Add const methods to access raw vector data. Non-constant vector data getters are syncing data between host and device as needed. While convenient in some instances, these methods are changing member data in `Vector`, so calling them prevents passing cont references to the `Vector` object.

We should probably do away with "sloppy" syncing altogether. This PR is a first step to enable using const references to `Vector` objects.

 
 
  
 

 ## Proposed changes
 
 

Implements following methods to `vector::Vector` class:
```c++
      const real_type* getData(MemorySpace memspace) const;
      const real_type* getData(index_type i, MemorySpace memspace) const;
```
 
 ## Checklist
 
 
 

     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 ## Further comments

### DIFF
--- a/resolve/vector/Vector.hpp
+++ b/resolve/vector/Vector.hpp
@@ -39,6 +39,8 @@ namespace ReSolve
       int        copyDataFrom(Vector* v, memory::MemorySpace memspaceIn, memory::MemorySpace memspaceOut);
       real_type* getData(memory::MemorySpace memspace);
       real_type* getData(index_type i, memory::MemorySpace memspace);
+      const real_type* getData(memory::MemorySpace memspace) const;
+      const real_type* getData(index_type i, memory::MemorySpace memspace) const;
 
       index_type getCapacity() const;
       index_type getSize() const;
--- a/resolve/vector/Vector.hpp
+++ b/resolve/vector/Vector.hpp
@@ -35,8 +35,8 @@ namespace ReSolve
       Vector(index_type n, index_type k);
       ~Vector();
 
-      int              copyDataFrom(const real_type* data, memory::MemorySpace memspaceIn, memory::MemorySpace memspaceOut);
-      int              copyDataFrom(Vector* v, memory::MemorySpace memspaceIn, memory::MemorySpace memspaceOut);
+      int              copyDataFrom(const real_type* source, memory::MemorySpace memspaceIn, memory::MemorySpace memspaceOut);
+      int              copyDataFrom(Vector* source, memory::MemorySpace memspaceIn, memory::MemorySpace memspaceOut);
       real_type*       getData(memory::MemorySpace memspace);
       real_type*       getData(index_type i, memory::MemorySpace memspace);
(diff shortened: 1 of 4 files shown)


alexander-novo:
You actually read my mind on this - I was going to ask you about this exact problem.

resolve/vector/Vector.hpp
```diff
@@ -35,10 +35,12 @@ namespace ReSolve
       Vector(index_type n, index_type k);
       ~Vector();
 
-      int        copyDataFrom(const real_type* data, memory::MemorySpace memspaceIn, memory::MemorySpace memspaceOut);
-      int        copyDataFrom(Vector* v, memory::MemorySpace memspaceIn, memory::MemorySpace memspaceOut);
-      real_type* getData(memory::MemorySpace memspace);
-      real_type* getData(index_type i, memory::MemorySpace memspace);
+      int              copyDataFrom(const real_type* data, memory::MemorySpace memspaceIn, memory::MemorySpace memspaceOut);
+      int              copyDataFrom(Vector* v, memory::MemorySpace memspaceIn, memory::MemorySpace memspaceOut);
```
alexander-novo: It should now be possible to have a version of `copyDataFrom` that takes a `const Vector*`

alexander-novo (approved):

pelesh:
> You actually read my mind on this - I was going to ask you about this exact problem.

It's been on our to-do list for a while. Thanks for your initiative!

resolve/vector/Vector.hpp
pelesh: `copyDataFrom` should in principle _only_ take `const Vector*` as the argument. We experimented with implicit data movement between host and device in an attempt to create nicer API but that didn't quite work out as we hoped for.

shakedregev: I would kind of like these functions to be named `copyDataFromTo` to indicate the argument order. Or if this is really objectionable, let's just drop `From`. Alone it gives no useful information.

shakedregev (commented):
Looks good, testing before approving.

shakedregev (approved):
Tests pass. Approving pending my minor tweak suggestions and a Changelog modification

resolve/vector/Vector.hpp
pelesh: The naming convention goes as follows:
- `copyDataFrom(Vector* source, ...)` means data is copied _from_ vector pointed by `source` _to_ `this`.
- `copyDataTo(real_type* dest, ...)` means data is copied _from_ `this` _to_ the data array pointed by `dest`.

Here `this` is the pointer to the instance implementing copy methods and it is not in the method argument list.

Using `const` keyword and naming pointers `source` and `destination` should improve readability of the code.

PR #409, opened 2026-02-23
Add 3 new element-wise vector operations

alexander-novo:
## Description
 
Add a new vector operations:
- `elementwiseDivide` which performs like `scale`, but scales by the inverse of a diagonal matrix.
- `elementwiseMax` which computes a vector whose elements are the maximum of the corresponding elements in the two input vectors.
- `abs` which computes a vector whose elements are the absolute values of the corresponding element in the input vector.
 

 ## Proposed changes
 
N/A
 
 ## Checklist

- [ ] All tests pass (`make test` and `make test_install` per testing [instructions](https://github.com/ORNL/ReSolve?tab=readme-ov-file#test-and-deploy)). Code tested on
     - [x] CPU backend
     - [ ] CUDA backend
     - [ ] HIP backend
- [ ] I have manually run the non-experimental examples and verified that residuals are close to machine precision. (In your build directory run:
`./examples/<your_example>.exe -h` to get instructions how to run examples). Code tested on:
     - [x] CPU backend
     - [ ] CUDA backend
     - [ ] HIP backend
- [x] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [x] The new code follows Re::Solve style guidelines.
- [x] There are unit tests for the new code.
- [x] The new code is documented.
- [x] The feature branch is rebased with respect to the target branch.
- [x] I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR. If this is a minor PR that is part of a larger fix already included in the file, state so.
 
 ## Further comments
 
N/A

PR #413, opened 2026-02-26
Add 3 new element-wise vector operations 

alexander-novo:
## Description
 
Add a new vector operations:
- `scaleInv` which performs like `scale`, but scales by the inverse of a diagonal matrix.
- `max` which computes a vector whose elements are the maximum of the corresponding elements in the two input vectors.
- `abs` which computes a vector whose elements are the absolute values of the corresponding element in the input vector.
 
Supplants #409 and is based on top of #411

 ## Proposed changes
 
N/A
 
 ## Checklist

     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 ## Further comments
 
N/A

### DIFF
--- a/resolve/cuda/cudaVectorKernels.cu
+++ b/resolve/cuda/cudaVectorKernels.cu
@@ -84,17 +84,17 @@ namespace ReSolve
       }
 
       /**
-       * @brief Divides a vector's elements by another vector's elements
+       * @brief Multiplies vector by an inverse of a diagonal matrix.
        *
        * @param[in]  n       - size of the vectors
-       * @param[in]  d_val   - divisor values
+       * @param[in]  d_val   - diagonal matrix values
        * @param[in, out] vec - vector to be divided. Changes in place.
        *
        * @todo Decide how to allow user to configure grid and block sizes.
        */
-      __global__ void scaleInv(index_type       n,
-                               const real_type* d_val,
-                               real_type*       vec)
+      __global__ void diagSolve(index_type       n,
+                                const real_type* d_val,
+                                real_type*       vec)
       {
         // Get the index of the element to be processed
         index_type idx = blockIdx.x * blockDim.x + threadIdx.x;
@@ -102,7 +102,7 @@ namespace ReSolve
         // Check if the index is within bounds
         if (idx < n)
         {
-          // Divide the vector element by the corresponding divisor value
+          // Divide the vector element by the corresponding diag value
           vec[idx] /= d_val[idx];
         }
       }
@@ -200,18 +200,18 @@ namespace ReSolve
      * @brief Wrapper that divides a vector's elements by another vector's elements
      *
      * @param[in]  n       - size of the vectors
-     * @param[in]  divisor - divisor values
+     * @param[in]  diag - diag values
      * @param[in, out] vec - vector to be divided. Changes in place.
      *
      * @todo Decide how to allow user to configure grid and block sizes.
      */
-    void scaleInv(index_type       n,
-                  const real_type* divisor,
-                  real_type*       vec)
+    void diagSolve(index_type       n,
+                   const real_type* diag,
+                   real_type*       vec)
     {
       int num_blocks = (n + block_size - 1) / block_size;
       // Launch the kernel
-      kernels::scaleInv<<<num_blocks, block_size>>>(n, divisor, vec);
+      kernels::diagSolve<<<num_blocks, block_size>>>(n, diag, vec);
     }
 
     /**
--- a/resolve/cuda/cudaVectorKernels.cu
+++ b/resolve/cuda/cudaVectorKernels.cu
@@ -90,9 +90,9 @@ namespace ReSolve
        *
        * @todo Decide how to allow user to configure grid and block sizes.
        */
-      __global__ void elementwiseDivide(index_type       n,
-                                        const real_type* d_val,
-                                        real_type*       vec)
+      __global__ void scaleInv(index_type       n,
+                               const real_type* d_val,
+                               real_type*       vec)
       {
         // Get the index of the element to be processed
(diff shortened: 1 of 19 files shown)


pelesh:
If we are using BLAS-like naming scheme as in #411, I suggest we rename the methods as follows:
```c++
void scaleInv(const Vector* D, Vector* x); // multiply each x[i] with 1./D[i]
void max(const Vector* x, const Vector* y, Vector* z); // z[i] = max(x[i], y[i])
void abs(const Vector* x, Vector* y); // y[i] = abs(x[i])
```
I would allow in `max`, for example, to have `x==z` or `y==z` for more flexibility. 

I also wouldn't hard-wire block size within kernels. Try something like
```c++
constexpr index_type block_size = 256;
```
within the proper namespace.

alexander-novo:
> I also wouldn't hard-wire block size within kernels. Try something like
> 
> ```c++
> constexpr index_type block_size = 256;
> ```
> 
> within the proper namespace.

The rest of the kernels in this module have hard-wired block sizes.

alexander-novo:
Should be ready to review

pelesh (approved):
I fixed issues with HIP and CUDA kernels. Would like review by @shakedregev before merging.

resolve/cuda/cudaVectorKernels.cu
```diff
@@ -81,8 +83,83 @@ namespace ReSolve
...
+       *
+       * @param[in]  n       - size of the vectors
+       * @param[in]  d_val   - divisor values
+       * @param[in, out] vec - vector to be divided. Changes in place.
+       *
+       * @todo Decide how to allow user to configure grid and block sizes.
+       */
+      __global__ void scaleInv(index_type       n,
+                               const real_type* d_val,
+                               real_type*       vec)
```
shakedregev: The distinction between the two roles should be obvious based on their names. Add a comment to how this relates to matrix solves ($`Ax=b \rightarrow b./A`$)
```suggestion
      __global__ void diagSolve(index_type       n,
                               const real_type* diagonal,
                               real_type*       vec)
```
update the documentation to clarify that this is a diagonal matrix. Update the name everywhere else where applicable.

resolve/cuda/cudaVectorKernels.cu
```diff
@@ -81,8 +83,83 @@ namespace ReSolve
...
+       * @param[in]  x   - First vector values
+       * @param[in]  y   - Second vector values
+       * @param[out] out - Output vector values
+       *
+       * @todo Decide how to allow user to configure grid and block sizes.
+       */
+      __global__ void max(index_type       n,
+                          const real_type* x,
+                          const real_type* y,
+                          real_type*       out)
```
shakedregev, line 123:
```suggestion
      __global__ void elementwiseMax(index_type       n,
                          const real_type* vec1,
                          const real_type* vec2,
                          real_type*       max)
```
update names everywhere

resolve/cuda/cudaVectorKernels.cu
```diff
@@ -81,8 +83,83 @@ namespace ReSolve
...
+       *
+       * @param[in]  n   - size of the vector
+       * @param[in]  in  - Vector input
+       * @param[out] out - Vector output
+       *
+       * @todo Decide how to allow user to configure grid and block sizes.
+       */
+      __global__ void abs(index_type       n,
+                          const real_type* in,
+                          real_type*       out)
```
shakedregev, line 147:
```suggestion
      __global__ void elementwiseAbs(index_type       n,
                          const real_type* in,
                          real_type*       out)
```

tests/unit/vector/VectorHandlerTests.hpp
```diff
@@ -280,6 +280,163 @@ namespace ReSolve
...
+        z.allocate(memspace_);
+
+        auto x_data = std::unique_ptr<real_type[]>(new real_type[N]);
+        auto y_data = std::unique_ptr<real_type[]>(new real_type[N]);
+        for (size_t i = 0; i < static_cast<size_t>(N); ++i)
+        {
+          if (i % 3 == 0)
+          {
+            x_data[i] = (real_type) (i + 1);
+            y_data[i] = (real_type) i * .5;
```
shakedregev:
```suggestion
            y_data[i] = (real_type) i * 0.5;
```

shakedregev (commented):
Fix the naming stuff, otherwise looks good. I still need to run it.

resolve/cuda/cudaVectorKernels.cu
alexander-novo: okay we need to figure out the naming scheme because this will be the 3rd name change, and it will be going back to what I just changed it from.

alexander-novo:
@shakedregev Can you show me where the floating point conventions are described? There are several other `.5` values without the `0` in the same file.

resolve/cuda/cudaVectorKernels.cu
alexander-novo: You also need to figure out the parameter naming scheme because I was going to do e.g. `vec1` and `vec2` but @pelesh asked for `x`, `y`, `z` etc.

shakedregev:
> @shakedregev Can you show me where the floating point conventions are described? There are several other `.5` values without the `0` in the same file.

You're correct that the rules don't say that. We will update them. Almost everywhere in the code, including in the file we wrote `0.5`, but yes it seems someone violated the unwritten convention there. The rationale is that you want it to be as obvious as possible that this is a number.

resolve/cuda/cudaVectorKernels.cu
pelesh: Changed in https://github.com/ORNL/ReSolve/pull/413/commits/5cbd6a4e81e1748e1b6c1edd3262815277bcb9f3

PR #416, opened 2026-02-27
Make Vector::copyDataTo able to copy from device to host and vice versa

### LINKED ISSUE #414: Make `Vector::copyDataTo` able to copy from device to host and vice versa
### Summary

Current implementation of `Vector::copyDataTo` methods allows for copying from host-to-host and device-to-device only. If one needs to copy from host to device one needs an additional copy operation. Need to modify `Vector::copyDataTo` so it can copy data from any memory space to any memory space, similar as `Vector::copyDataFrom` methods.

### Rationale

Eliminating additional data copy when using `Vector::copyDataTo` calls will reduce unnecessary performance overhead. This is also required if we want to eliminate all "sloppy" data movement that `Vector::getData` functions do. 

### Description

Implement methods:
```c++
      int copyDataTo(real_type* dest, index_type i, memory::MemorySpace memspaceSrc, memory::MemorySpace memspaceDst);
      int copyDataTo(real_type* dest, memory::MemorySpace memspaceSrc, memory::MemorySpace memspaceDst);
```
that can copy from any memory space (host or device) to any memory space (host or device).

### Additional information

_No response_

shakedregev:
## Description
 
Current implementation of `Vector::copyDataTo` methods allows for copying from host-to-host and device-to-device only. If one needs to copy from host to device one needs an additional copy operation. Need to modify `Vector::copyDataTo` so it can copy data from any memory space to any memory space, similar as `Vector::copyDataFrom` methods.
 
  

 ## Proposed changes
 
My changes allow the source and memory space to be different. I also addressed naming issues and syncing issues.
 
 ## Checklist

     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 ## Further comments
 
I noticed some issues that I left to not explode the scope of the PR. We should follow up on these.

1. `copyDataFrom` has similar naming and potentially syncing issues
2. GramSchmidt is littered with `deviceSynchronize()` seemingly arbitrarily
3. Vector tests have some commented tests.

### DIFF
--- a/resolve/GramSchmidt.cpp
+++ b/resolve/GramSchmidt.cpp
@@ -259,7 +259,7 @@ namespace ReSolve
         H[idxmap(i, j, num_vecs_ + 1)] -= s;
       } // for j
       vec_Hcolumn_->resize(i + 1);
-      vec_Hcolumn_->copyDataFrom(&H[idxmap(i, 0, num_vecs_ + 1)], memory::HOST, memspace_);
+      vec_Hcolumn_->copyFromExternal(&H[idxmap(i, 0, num_vecs_ + 1)], memory::HOST, memspace_);
       vector_handler_->massAxpy(n, vec_Hcolumn_, i + 1, V, vec_w_, memspace_);
 
       // normalize (second synch)
@@ -349,7 +349,7 @@ namespace ReSolve
       }
 
       vec_Hcolumn_->resize(i + 1);
-      vec_Hcolumn_->copyDataFrom(&H[idxmap(i, 0, num_vecs_ + 1)], memory::HOST, memspace_);
+      vec_Hcolumn_->copyFromExternal(&H[idxmap(i, 0, num_vecs_ + 1)], memory::HOST, memspace_);
 
       vector_handler_->massAxpy(n, vec_Hcolumn_, i + 1, V, vec_w_, memspace_);
       // normalize (second synch)
--- a/resolve/GramSchmidt.cpp
+++ b/resolve/GramSchmidt.cpp
@@ -187,7 +187,7 @@ namespace ReSolve
       // copy H_col to aux, we will need it later
       vec_Hcolumn_->setDataUpdated(memspace_);
       vec_Hcolumn_->resize(i + 1);
-      vec_Hcolumn_->copyDataTo(h_aux_, 0, memory::HOST);
+      vec_Hcolumn_->copyToExternal(h_aux_, 0, memory::HOST);
       mem_.deviceSynchronize();
 
       // Hcol = V(:,1:i)^T*V(:,i+1);
@@ -200,7 +200,7 @@ namespace ReSolve
 
       // copy H_col to H
       vec_Hcolumn_->setDataUpdated(memspace_);
-      vec_Hcolumn_->copyDataTo(&H[idxmap(i, 0, num_vecs_ + 1)], 0, memory::HOST);
+      vec_Hcolumn_->copyToExternal(&H[idxmap(i, 0, num_vecs_ + 1)], 0, memory::HOST);
       mem_.deviceSynchronize();
 
       // add both pieces together (unstable otherwise, careful here!!)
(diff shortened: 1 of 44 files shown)


alexander-novo:
with the recent change to `getData` being `const` is it possible to mark these functions as `const`?

resolve/vector/Vector.cpp
```diff
@@ -842,70 +842,149 @@ namespace ReSolve
...
      * @post All elements of the vector _i_ are copied to the array _dest_.
      */
-    int Vector::copyDataTo(real_type*          destination,
-                           index_type          i,
-                           memory::MemorySpace memspaceInOut)
+    int Vector::copyToExternal(real_type* dest, index_type i, memory::MemorySpace memspaceSrc, memory::MemorySpace memspaceDst)
     {
       using namespace ReSolve::memory;
+      real_type* data = this->getData(i, memspaceSrc);
+      // Check that the source data is not null and up to date
```
pelesh: I suggest we remove redundant `this`:
```suggestion
      real_type* data = getData(i, memspaceSrc);
      // Check that the source data is not null and up to date
```

resolve/GramSchmidt.cpp
```diff
@@ -200,7 +200,7 @@ namespace ReSolve
 
       // copy H_col to H
       vec_Hcolumn_->setDataUpdated(memspace_);
-      vec_Hcolumn_->copyDataTo(&H[idxmap(i, 0, num_vecs_ + 1)], 0, memory::HOST);
+      vec_Hcolumn_->copyToExternal(&H[idxmap(i, 0, num_vecs_ + 1)], 0, memory::HOST, memory::HOST);
       mem_.deviceSynchronize();
```
pelesh, line 204: Hessenberg matrix `H` should be defined as multivector instead of raw array. This is probably out of scope of this PR.

pelesh (commented):
I would suggest renaming `copyDataFrom` --> `copyFromExternal` in this PR. Otherwise we leave vector methods names in "asymmetric" state.

Other than that, this looks good!

pelesh:
> with the recent change to `getData` being `const` is it possible to mark these functions as `const`?

It is, but would require cleaning up the code where the sloppy data movement was used.

shakedregev:
Thanks for the feedback. Fixed. Please re-review.

resolve/hykkt/cholesky/CholeskySolverHip.cpp
```diff
@@ -169,7 +169,7 @@ namespace ReSolve
      */
     void CholeskySolverHip::solve(vector::Vector* x, vector::Vector* b)
     {
-      x->copyDataFrom(b, memory::DEVICE, memory::DEVICE);
+      x->copyFromExternal(b, memory::DEVICE, memory::DEVICE);
       // TODO: currently, this returns status rocblas_status_invalid_pointer
       // but we have verified that none of the inputs are null and need to be non-null
```
shakedregev:
```suggestion
// TODO: currently, this returns status rocblas_status_invalid_pointer
// but we have verified that none of the inputs are null and need to be non-null.
```

resolve/hykkt/spgemm/SpGEMMCpu.cpp
```diff
@@ -97,11 +97,11 @@ namespace ReSolve
 
       // Previous data must be de-allocated and new data copied
       // Cholmod does not allow for reuse of arrays
```
shakedregev:
```suggestion
// Previous data must be de-allocated and new data copied.
// Cholmod does not allow for reuse of arrays.
```

resolve/matrix/Coo.cpp
```diff
@@ -240,7 +240,7 @@ namespace ReSolve
     if (memspaceOut == memory::DEVICE)
     {
       // check if cuda data allocated
```
shakedregev:
```suggestion
      // Check if Cuda data is allocated.
```

pelesh (approved):

PR #417, opened 2026-03-04
Support left preconditioning in GMRES 

kakeueda:
## Description

 This change allows users to solve a left-preconditioned system $M^{-1}Ax = M^{-1}b$ using GMRES. It also adds a PreconditionerUserMatrix class.

@JeffZ594 My understanding is that we eventually want to add regularization to the GMRES solvers. We can follow up with another PR that adds regularization support with testTripsProblem.

 ## Proposed changes
 
- Updated GMRES to support left preconditioning.
- Added a PreconditionerUserMatrix class so users can define $M^{-1}$ explicitly.
- Added unit tests for lu and usermatrix preconditioners.
- Added a new command line option `-p left (or right)` for the GMRES solver (default: right).
 
 ## Checklist 

     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 ## Further comments
 
Non-flexible GMRES only. If FGMRES is selected with `-p left`, the solver reports an error.


JeffZ594:
@kakeueda 

Thank you for making the unit tests for the preconditioners! I have code that implements regularization in Eigen that needs to be translated that would be better for testTripsProblem. I think these tests work better for testing just the preconditioners.

kakeueda:
> @kakeueda
> 
> Thank you for making the unit tests for the preconditioners! I have code that implements regularization in Eigen that needs to be translated that would be better for testTripsProblem. I think these tests work better for testing just the preconditioners.

Sure, let me first finish this PR and I'll take a look at your implementation .
@pelesh Are you okay with adding Eigen as a new dependency?

JeffZ594:
@kakeueda I believe Dr Peles does not want to use Eigen as a dependency. Right now I only use Eigen to calculate the SVD of the Hessenberg Matrix for GMRES but I believe there are alternatives like LAPACK that can also do this.

pelesh:
> @kakeueda I believe Dr Peles does not want to use Eigen as a dependency. Right now I only use Eigen to calculate the SVD of the Hessenberg Matrix for GMRES but I believe there are alternatives like LAPACK that can also do this.

Hessenberg matrix is dense, so SVD from LAPACK should work in this case. I wouldn't add Eigen as a dependency unless there is compelling enough reason to do so.

kakeueda:
For now the default behavior for left preconditioning is to use the preconditioned residual $||M^{-1}(b - Ax)||$ for convergence check and the true residual $||(b - Ax)||$ for monitoring (final_residual_norm_ stores this).

kakeueda:
This is outside the scope of this PR but 'solve()' in GMRES is starting to get a bit cluttered. It might be worth considering adding a base GMRES class and having the current RandFGMRES and regularized GMRES derive from it. We could also clean up the convergence check for GMRES which I believe this was mentioned before.

resolve/LinSolverIterativeFGMRES.cpp
```diff
@@ -284,12 +347,18 @@ namespace ReSolve
...
+        if (preconditioner_->getSide() == "right")
+        {
+          // The correction is M^{-1}*vec_z for right preconditioning
+          preconditioner_->apply(&vec_z, &vec_v);
+          // Add the correction to x
+          vector_handler_->axpy(ONE, &vec_v, x, memspace_);
+        }
+        else
+        {
+          // Add the correction to x directry for left preconditioning
```
shakedregev:
```suggestion
          // Add the correction to x directly for left preconditioning.
```
Spelling correction.

resolve/LinSolverIterativeRandFGMRES.cpp
```diff
@@ -352,12 +413,18 @@ namespace ReSolve
...
+        if (preconditioner_->getSide() == "right")
+        {
+          // The correction is M^{-1}*vec_z for right preconditioning
+          preconditioner_->apply(&vec_z, &vec_v);
+          // Add the correction to x
+          vector_handler_->axpy(ONE, &vec_v, x, memspace_);
+        }
+        else
+        {
+          // Add the correction to x directry for left preconditioning
```
shakedregev:
```suggestion
          // Add the correction to x directly for left preconditioning.
```

resolve/LinSolverIterativeRandFGMRES.cpp
```diff
@@ -392,15 +466,16 @@ namespace ReSolve
 
       if (!outer_flag)
       {
-        rnorm = vector_handler_->dot(vec_V_, vec_V_, memspace_);
-        // rnorm = ||V_0||
-        rnorm = std::sqrt(rnorm);
+        // true rnorm = ||b - A*x||
+        true_rnorm = vector_handler_->dot(vec_V_, vec_V_, memspace_);
+        true_rnorm = std::sqrt(true_rnorm);
```
shakedregev: maybe we should change this to `true_res_norm`? `r` can be anything, it's not clear from context (and there's no separator from norm)

tests/unit/preconditioner/PreconditionerUserMatrixTests.hpp
```diff
@@ -0,0 +1,187 @@
...
+          status *= false;
+        }
+
+        return status.report(__func__);
+      }
+
+      /**
+       * @brief Test apply() performs correct matvec with B.
+       *
+       * B = diag(2, 3, 4), x = [1, 1, 1]^T => y = [2, 3, 4]^T.
```
shakedregev: Why tranpose?

shakedregev (commented):
Some minor comments. I need to test and I will change this to approve.

shakedregev (approved):
Great work.

resolve/LinSolverIterativeRandFGMRES.cpp
kakeueda: Agree, also changed bnorm to rhs_norm

tests/unit/preconditioner/PreconditionerUserMatrixTests.hpp
kakeueda: It was meant to show a column vector, but maybe unnecessary

kakeueda:
returned 503: Egress is over the account limit.

I got this error in the test. Any idea what might be causing this? @pelesh

nkoukpaizan:
> returned 503: Egress is over the account limit.
> 
> I got this error in the test. Any idea what might be causing this? @pelesh

Looks like a temporary failure in GH-Actions. It's working now!

tests/unit/preconditioner/PreconditionerUserMatrixTests.hpp
shakedregev: Worse, it's misleading. Vectors are by default column, unless you put a transpose. That's what didn't make sense to me. I knew it had to be a column, but this is the convention for row.

resolve/LinSolverIterativeFGMRES.cpp
```diff
@@ -130,21 +139,51 @@ namespace ReSolve
...
+      vec_v.setData(rhs->getData(memspace_), memspace_);
+      preconditioner_->apply(&vec_v, &vec_z);
+      rhs_norm = vector_handler_->dot(&vec_z, &vec_z, memspace_);
+      rhs_norm = std::sqrt(rhs_norm);
+    }
+    else
+    {
+      res_norm = true_res_norm;
+      rhs_norm = true_rhs_norm;
+    }
```
pelesh: Why is left preconditioner not applied to flexible GMRES?

resolve/LinSolverIterativeFGMRES.cpp
```diff
@@ -130,21 +139,51 @@ namespace ReSolve
...
+    {
+      // Left-preconditioned residual norm ||M^{-1}*(b - A*x0)||
+      preconditioner_->apply(&vec_v, &vec_z);
+      vec_v.copyFromExternal(&vec_z, memspace_, memspace_);
+      res_norm = vector_handler_->dot(vec_V_, vec_V_, memspace_);
+      res_norm = std::sqrt(res_norm);
+
+      // Left-preconditioned right-hand side norm ||M^{-1}*b||
+      vec_v.setData(rhs->getData(memspace_), memspace_);
+      preconditioner_->apply(&vec_v, &vec_z);
```
pelesh: Why use `setData` function here? This method will be removed and should not be used in new contributions.

resolve/LinSolverIterativeFGMRES.cpp
```diff
@@ -284,37 +347,57 @@ namespace ReSolve
...
           vector_handler_->axpy(h_rs_[j], &vec_v, &vec_z, memspace_);
         }
-        // now multiply d_Z by precon
-
-        vec_v.setData(vec_V_->getData(memspace_), memspace_);
-        this->precV(&vec_z, &vec_v);
-        // and add to x
-        vector_handler_->axpy(ONE, &vec_v, x, memspace_);
+        if (preconditioner_->getSide() == "right")
+        {
```
pelesh: It is safer to use here enums instead of strings.

resolve/LinSolverIterativeRandFGMRES.cpp
```diff
@@ -147,9 +154,11 @@ namespace ReSolve
...
     int        k1;
 
-    real_type   t;
-    real_type   rnorm;
-    real_type   bnorm;
+    real_type   t             = 0.0;
+    real_type   res_norm      = 0.0;
+    real_type   rhs_norm      = 0.0;
+    real_type   true_res_norm = 0.0;
+    real_type   true_rhs_norm = 0.0;
```
pelesh: What is "true" residual norm? This needs to be documented in the code.

resolve/LinSolverIterativeRandFGMRES.cpp
```diff
@@ -244,14 +280,39 @@ namespace ReSolve
...
+          mem_.deviceSynchronize();
 
-        matrix_handler_->matvec(A_, &vec_z, &vec_v, &ONE, &ZERO, memspace_);
+          vec_v.setData(vec_V_->getData(i + 1, memspace_), memspace_);
+          matrix_handler_->matvec(A_, &vec_z, &vec_v, &ONE, &ZERO, memspace_);
+        }
+        else
+        {
+          if (preconditioner_->getSide() == "right")
+          {
```
pelesh: I suggest using sitch statement and enums instead of strings and if-else statements here.

resolve/LinSolverIterativeRandFGMRES.cpp
```diff
@@ -352,16 +413,22 @@ namespace ReSolve
...
+        {
+          // The correction is M^{-1}*vec_z for right preconditioning
+          preconditioner_->apply(&vec_z, &vec_v);
+          // Add the correction to x
+          vector_handler_->axpy(ONE, &vec_v, x, memspace_);
+        }
+        else
+        {
+          // Add the correction to x directly for left preconditioning.
+          vector_handler_->axpy(ONE, &vec_z, x, memspace_);
```
pelesh: The comments in the code need to be more specific explaining steps clearly.

resolve/LinSolverIterativeRandFGMRES.cpp
```diff
@@ -385,22 +459,23 @@ namespace ReSolve
...
+        true_res_norm = std::sqrt(true_res_norm);
 
         io::Logger::misc() << "End of cycle, COMPUTED norm of residual "
                            << std::scientific << std::setprecision(16)
-                           << rnorm << "\n";
+                           << true_res_norm << "\n";
 
-        final_residual_norm_ = rnorm / bnorm; // relative residual norm
+        // Report the true relative residual norm
+        final_residual_norm_ = true_res_norm / true_rhs_norm;
```
pelesh, line 496: Really need to explain what is "true" residual norm.

tests/functionality/CMakeLists.txt
```diff
@@ -182,6 +182,22 @@ add_test(NAME sys_gmres_mgspm_test
...
+                 "randgmres" "-g" "mgs" "-s" "count" "-p" "left"
+)
+add_test(NAME sys_gmres_left_cgs2_test
+         COMMAND $<TARGET_FILE:sys_rand_gmres_test.exe> "-x" "no" "-i" "fgmres"
+                 "-g" "cgs2" "-p" "left"
+)
+add_test(NAME sys_gmres_left_mgs_test
+         COMMAND $<TARGET_FILE:sys_rand_gmres_test.exe> "-x" "no" "-i" "fgmres"
+                 "-g" "mgs" "-p" "left"
+)
```
pelesh, line 226: We should consider adding more combinations of different preconditioners and other solver options.

pelesh (changes requested):
Very nice work and good testing coverage of the new preconditioner features. A few things that need to be done before we merge this:
- The GMRES code needs to be better commented to explain the workflow.
- Use enums instead of strings to select preconditioner type.
- There should be a warning message if user tries to use e.g. FGMRES with unsupported preconditioner and the preconditioner choice should fall back to default.

Also see some more specific comments below.

resolve/LinSolverIterativeFGMRES.cpp
kakeueda: FGMRES is generally defined as a right-preconditioned method. Please see Section 9.4 (https://www.mat.uc.pt/~alma/aulas/NewTrends/Books/Saad.pd). One difficulty with left preconditioning in the flexible setting is that the residual $M^{-1}(Ax-b)$ can change at each iteration as the preconditioner changes. petsc also supports only right preconditioning for FGMRES.

resolve/LinSolverIterativeRandFGMRES.cpp
kakeueda: Added documentation explaining what the true residual is. This separates the residual used for convergence checking from the residual reported to the user. For right preconditioning the true residual $Ax - b$ is used for both, but for left preconditioning the preconditioned residual $M^{-1}(Ax - b)$ is used for convergence checking, while $Ax - b$ is still reported so users can monitor the residual of the original system $Ax = b$. We could also add an option to report the preconditioned residual but not the scope of this PR because TestHelper would also need to be updated to compute the preconditioned residual.

kakeueda:
@pelesh Thank you for your suggestions. 
1. Replaced std::string with an enum for the preconditioning side
2. Improved the comments explaining the workflow and residuals 
3. More tests are added for left preconditioning
4. Changed the behavior to fall back to the default option (right) instead of throwing an error when FGMRES and left preconditioning are selected.

resolve/LinSolverIterativeFGMRES.cpp
kakeueda: Is there currently an equivalent to setData that we should use instead?
I thought we would still need setData because vec_v should be a view into the multivector vec_V_, not a copy.

shakedregev:
@pelesh - can you confirm that your concerns are addressed so we can merge this?

shakedregev:
@JeffZ594 - It would be better if a user uses an incompatible preconditioner with FGMRES to throw an error. An incompatible preconditioner and solver likely means an error in preconditioner construction or calling the wrong preconditioner or solver. Can you fix this?

shakedregev (changes requested):
Please ensure that the preconditioner and solver are compatible. If not, throw an error.
Do not assume a default right preconditioner. The options should be `left`, `right` or throw an error.

Having defaults is good, to not overburden the user. But if the user specifies an input, and it's invalid, that usually means an error in the logic.

kakeueda:
> Please ensure that the preconditioner and solver are compatible. If not, throw an error. Do not assume a default right preconditioner. The options should be `left`, `right` or throw an error.
> 
> Having defaults is good, to not overburden the user. But if the user specifies an input, and it's invalid, that usually means an error in the logic.

Yes, I also think the solver should not run against the user's input. Changed the behavior to throw an error when the solver and preconditioner are incompatible.

resolve/LinSolverIterativeFGMRES.cpp
pelesh: It is not clear to me why you need `vec_v` to begin with. Can't you just use
```c++
preconditioner_->apply(&rhs, &vec_z);
```
?

resolve/LinSolverIterativeRandFGMRES.cpp
pelesh: In this case, if the string says "middle", it will still be interpreted as left preconditioner.

resolve/LinSolverIterativeFGMRES.cpp
kakeueda: You’re right! It was redundant to begin with

resolve/LinSolverIterativeRandFGMRES.cpp
kakeueda: This has already been addressed in the latest commit using a switch statement. If an user uses anything other than "right" or "left", it'll throw an error

resolve/LinSolverIterativeFGMRES.cpp
kakeueda: All tests passed without `vec_v`.

tests/functionality/CMakeLists.txt
pelesh: I guess we can merge without this requirement, but we do need better test coverage.

JeffZ594: I believe that the tests with the solver options are comprehensive for the default preconditioner being used from the left. I can add an option to use $A^T$ as a preconditioner within sys_rand_gmres_test.exe to test the user defined preconditioner. @pelesh were you looking for something like this or a brand new test?

pelesh: Eventually we need to test every single combination of solver, GS and preconditioner. I would leave that part out of this PR.

pelesh:
I think this is ready to merge. Let's rebase and re-run all tests.

shakedregev (approved):
All test pass except Cholesky on HIP. This is a known bug that will be fixed after merging #421.

PR #421, opened 2026-05-15
Andrew/sccg and matrix class

andrewxu319:
## Description
 
 
 
 
 
  

Implemented Schur complement conjugate gradient and Matrix parent class.

 ## Proposed changes
 
Added a CPU implementation of Schur complement conjugate gradient similar to the [HyKKT implementation](https://github.com/ORNL/hykkt/blob/develop/src/SchurComplementConjugateGradient.cpp), along with a basic test that will need to be completed later. The current test is a modified copy of the tests in the HyKKT implementation, which uses a set of test matrix files.

Created a Matrix class that is a parent of Sparse and, in the future, other matrix representations. The Matrix class is designed to act as an interface for matrices generally and has wrappers for matrix operations implemented in MatrixHandler.

@shakedregev 
 
 
 ## Checklist
 
 
 

     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 ## Further comments


pelesh (dismissed):
Very nice work, thank you. Please remove `matrix::Matrix` class so we can merge this.

pelesh:
Some tests failing, need to fix that, as well. Setting this to draft PR until blocker issues are resolved.

tests/unit/hykkt/HykktSCCGTests.hpp
```diff
@@ -0,0 +1,150 @@
...
+        sccg.addVectorInfo(x0, b);
+        sccg.setup();
+        sccg.solve();
+
+        // // //
+        // real_type* x0_output_data = x0.getData(0, memspace_);
+        // for (index_type i = 0; i < n; i++) {
+        //   std::cout << x0_output_data[i] << ", ";
+        // }
+        // // //
```
shakedregev:
```suggestion
```

shakedregev (commented):
Agree with Slaven's comments. Make sure all previous tests pass. 
Slaven and I discussed yesterday and decided we don't need a matrix parent class. Sorry for the confusion.
Please pull my change before continuing.

andrewxu319:
Done--removed the Matrix class, and it's passing all of the tests.

resolve/hykkt/sccg/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+
+target_include_directories(
+  resolve_hykkt_spgemm PUBLIC ${SUITESPARSE_INCLUDE_DIR}
+)
+
+if(RESOLVE_USE_CUDA)
+  target_sources(resolve_hykkt_sccg PRIVATE ${Matrix_CUDASDK_SRC})
+  target_link_libraries(resolve_hykkt_sccg PUBLIC resolve_backend_cuda)
+endif()
+
```
shakedregev, line 22:
```suggestion

# Link to HIP ReSolve backend if HIP is support enabled

```

shakedregev:
> Done--removed the Matrix class, and it's passing all of the tests.

Good work on this. I made a small correction. I tested the codes on CUDA and HIP and the good news is that at least the old tests are still passsing. Before we merge this, can you do the HIP/CUDA implementation as well? It's not working currently.

shakedregev (approved):
Good work on this, I think it's ready for merging.

resolve/matrix/io.cpp
```diff
@@ -736,7 +736,13 @@ namespace ReSolve
...
     {
       std::list<MatrixElementTriplet>::iterator it = tmp.begin();
-      while (it != tmp.end())
+
+      if (it == tmp.end())
+      {
+        return 0;
+      }
+
+      while (it != std::prev(tmp.end()))
```
pelesh, line 747: Put comments to explain what is going on here. This code looks a little cryptic.

resolve/matrix/MatrixHandlerCuda.cpp
```diff
@@ -76,17 +76,23 @@ namespace ReSolve
     // result = alpha *A*x + beta * result
     cusparseStatus_t     status;
     cusparseDnVecDescr_t vecx = workspace_->getVecX();
-    cusparseCreateDnVec(&vecx, A->getNumRows(), vec_x->getData(memory::DEVICE), CUDA_R_64F);
+    cusparseCreateDnVec(&vecx, A->getNumColumns(), vec_x->getData(memory::DEVICE), CUDA_R_64F);
```
pelesh, line 85: This is pretty big change. Needs explanation and comments in the code.

resolve/matrix/MatrixHandlerCuda.cpp
```diff
@@ -76,17 +76,23 @@ namespace ReSolve
...
 
-    cusparseSpMatDescr_t matA = workspace_->getSpmvMatrixDescriptor();
-
-    void*            buffer_spmv     = workspace_->getSpmvBuffer();
     cusparseHandle_t handle_cusparse = workspace_->getCusparseHandle();
-    if (values_changed_)
+    // Rebuild cached SpMV setup if the matrix object or dimensions changed.
+    bool matrix_changed = (matrix_for_matvec_ != A) || (matvec_num_rows_ != A->getNumRows()) || (matvec_num_cols_ != A->getNumColumns()) || (matvec_nnz_ != A->getNnz());
+    if (matrix_changed || values_changed_)
+    {
```
pelesh: Break this line per style guidelines.

resolve/matrix/MatrixHandlerCuda.hpp
```diff
@@ -56,7 +56,12 @@ namespace ReSolve
 
   private:
     LinAlgWorkspaceCUDA* workspace_{nullptr};
-    bool                 values_changed_{true}; ///< needed for matvec
+    bool                 values_changed_{true}; ///< Flag to indicate if matrix values changed since the cached SpMV setup.
+
+    matrix::Sparse* matrix_for_matvec_{nullptr}; ///< Matrix used for cached SpMV setup.
+    index_type      matvec_num_rows_{0};         ///< Number of rows in cached SpMV matrix.
+    index_type      matvec_num_cols_{0};         ///< Number of columns in cached SpMV matrix.
+    index_type      matvec_nnz_{0};              ///< Number of nonzeros in cached SpMV matrix.
```
pelesh, line 67: Why is matrix cached to begin with?

resolve/matrix/MatrixHandlerHip.cpp
```diff
@@ -78,6 +78,13 @@ namespace ReSolve
 
     rocsparse_handle handle_rocsparse = workspace_->getRocsparseHandle();
 
+    // Rebuild cached SpMV setup if the matrix object or dimensions changed.
+    bool matrix_changed = (matrix_for_matvec_ != A) || (matvec_num_rows_ != A->getNumRows()) || (matvec_num_cols_ != A->getNumColumns()) || (matvec_nnz_ != A->getNnz());
+    if (matrix_changed || values_changed_)
+    {
+      workspace_->resetMatvecSetup();
+    }
```
pelesh: Break the long line.

tests/unit/hykkt/HykktSCCGTests.hpp
```diff
@@ -0,0 +1,170 @@
...
+       * The test fixture uses caller-provided matrix and vector handlers so the same test can be run with CPU, CUDA, or HIP backends.
+       *
+       * @param[in] memspace Memory space for the test (HOST or DEVICE).
+       * @param[in] matrix_handler Reference to a matrix handler for the selected backend.
+       * @param[in] vector_handler Reference to a vector handler for the selected backend.
+       * @param[in] generator Reference to a C++ random number generator.
+       */
+      HykktSchurComplementConjugateGradientTests(memory::MemorySpace memspace, MatrixHandler& matrix_handler, VectorHandler& vector_handler, std::mt19937& generator)
+        : memspace_(memspace), matrix_handler_(matrix_handler), vector_handler_(vector_handler), generator_(generator)
+      {
```
pelesh: Break long lines per style guidelines.

pelesh (dismissed):
Good work, some minor documentation (code commenting) and style issues need to be resolved before merging.

resolve/matrix/MatrixHandlerCuda.cpp
tamar-dewilde: This change comes from the dimensions of SpMV. If A is m x n, then in y = A*x, x has length n, and y has length m, so the input vector descriptor should use the matrix column count, while the output vector descriptor should use the row count. In the SCCG path I observed that JC and JC^T can be rectangular, so using `getNumRows()` for the input vector only works when the matrix happens to be square. Adjusted comments in code to clarify.

resolve/matrix/MatrixHandlerCuda.hpp
tamar-dewilde: SpMV itself does not need caching to work. Another option would be to rebuild the setup every time `matvec()` is called. I added the matrix tracking because the GPU workspace reuses the SpMV setup between calls. Since SCCG can use different matrices like JC and JC^T, the cached setup may not match the next matrix passed into `matvec()`. The tracking lets the handler detect that and reset the setup when the matrix changes. So the cache is mainly to avoid extra setup work, and the tracking makes that reuse safe. (edited for clarity after rereading an area I misinterpreted let me know if this implementation should be changed)

tests/unit/hykkt/HykktSCCGTests.hpp
andrewxu319: All requested changes completed.

PR #426, opened 2026-05-29
Removed unnecessary syncs and handles

shakedregev:
The code was synchronizing unnecessarily in GramSchmidt and FGMRES, no matter the backend. This should be handled on the HIP backend (which it now is). Removed double synching and all syncing from generic and CUDA code. Established a convention that syncing is always post operation (so there is no need to sync pre, because the device is synced from the post of the last operation).

 ## Proposed changes
 
Removed unnecessary syncs. Updates changelogs.
 
 ## Checklist
 
 
 

- [ ] All tests pass (`make test` and `make test_install` per testing [instructions](https://github.com/ORNL/ReSolve?tab=readme-ov-file#test-and-deploy)). Code tested on
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
- [ ] I have manually run the non-experimental examples and verified that residuals are close to machine precision. (In your build directory run: `./examples/<your_example>.exe -h` to get instructions how to run examples). Code tested on:
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 
 - Added `cons` counterparts to `Vector::getData` methods.
 
-- Made Vector::copyDataTo able to copy from device to host and vice versa
+- Made Vector::copyToExternal able to copy from device to host and vice versa
 
 - Added `diagSolve`, `max`, and `abs` vector operations.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,11 +16,15 @@
 
 - Added `cons` counterparts to `Vector::getData` methods.
 
-- Made Vector::copyToExternal able to copy from device to host and vice versa
+- Made Vector::copyToExternal able to copy from device to host and vice versa.
 
 - Added `diagSolve`, `max`, and `abs` vector operations.
 
 - Improved coding guidelines for developers on floating point conventions.
+
+- Removed unnecessary device synchronization.
+
+- Changed variable and function names to be more explanatory.
 
 ## Changes to Re::Solve in release 0.99.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - Removed unnecessary device synchronization and added where needed (HIP only calls).
 
 - Changed variable and function names to be more explanatory.
+
 - Added Developer Guide documentation for vector and matrix classes and handlers.
 
 ## Changes to Re::Solve in release 0.99.2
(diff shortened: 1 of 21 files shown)


pelesh (commented):
I believe these syncs are needed when using HIP.

shakedregev:
> I believe these syncs are needed when using HIP.

@pelesh - they are not. The code works without them. They are needed within the HIP specific code, but there is no need for them outside. We've added them in the HIP specific code.

pelesh:
> > I believe these syncs are needed when using HIP.
> 
> @pelesh - they are not. The code works without them. They are needed within the HIP specific code, but there is no need for them outside. We've added them in the HIP specific code.

Sorry, where in the HIP specific code are sync statements? I looked at `gemv` function in `VectorHandlerHip` and there is no sync call.

CHANGELOG.md
```diff
@@ -16,12 +16,15 @@
...
 - Added `cons` counterparts to `Vector::getData` methods.
 
-- Made Vector::copyDataTo able to copy from device to host and vice versa
+- Made Vector::copyToExternal able to copy from device to host and vice versa.
 
 - Added `diagSolve`, `max`, and `abs` vector operations.
 
 - Improved coding guidelines for developers on floating point conventions.
 
+- Removed unnecessary device synchronization.
```
shakedregev:
```suggestion
- Removed unnecessary device synchronization and added where needed (HIP only calls).
```

shakedregev:
@pelesh - fixed and I went ahead and fixed it in other places too. Tested on all platforms.

CHANGELOG.md
```diff
@@ -16,12 +16,15 @@
...
-- Made Vector::copyDataTo able to copy from device to host and vice versa
+- Made Vector::copyToExternal able to copy from device to host and vice versa.
 
 - Added `diagSolve`, `max`, and `abs` vector operations.
 
 - Improved coding guidelines for developers on floating point conventions.
 
+- Removed unnecessary device synchronization and added where needed (HIP only calls).
+
+- Changed variable and function names to be more explanatory.
```
shakedregev, line 27:
```suggestion
- Removed unnecessary device synchronization and added where needed (HIP only calls).

- Changed variable and function names to be more explanatory.

```

andrewxu319:
Suggested: synchronize at VectorHandlerHip.cpp, line 192. Premature return skips the sync at the bottom of the function.

andrewxu319 (approved):
Looks good, tests passing on my end too.

PR #427, opened 2026-06-01
Add initial guess support for iterative solvers

### LINKED ISSUE #376: Incorporate initial guesses into residual for iterative solvers.
### Summary

We should have an option to add an initial guess (for iterative refinement or regular iterative solve).
If there is an initial guess $`x`$ to the solution of $`Ax=b`$, we can compute $`r=b-Ax`$.
We then solve $`Ay=r`$ and return $`x+y`$ as the solution.
An ideal implementation is one where we have the same function name and it can take either two arguments ($`A,b`$) or three ($`A,b,x`$). The first is the current implementation, and the second will add a wrapper that stores the initial guess, and add it to the solution obtained from solving $`Ay=r`$.

### Rationale

Right now, when we use an iterative solver directly, we have no option of supplying an initial guess. When we use it for iterative refinement, we use an initial guess in a way that is less numerically accurate.

### Description

Some features to make this complete:

1. Check if $`|b|_2`<`|b-Ax|_2`$. If it is, warn the user that their initial guess is worse than all 0s and ignore that initial guess.
2. If a nonzero guess was supplied, check if $`|b-Ax-Ay|_2`>`|b-Ax|_2`$. Note that the order of subtraction is important for the left side. If it is, warn the user that the iterative solver was not able to find a solution better than the initial guess, and output $`x`$ (not $`x+y`$). For iterative refinement, this corresponds to not improving upon the direct solve.

### Additional information

_No response_

tamar-dewilde:
## Description

The current solver already uses `solve(rhs, x)` for the initial guess path. This PR adds the remaining safeguards requested in #376 and adds tests for the bad guess and accepted guess cases.
 
Closes #376 
 
@shakedregev 

 ## Proposed changes

- Warn and ignore the initial guess if its residual is worse than the zero-vector residual.
- Restore the accepted initial guess if the correction does not improve it.
- Preserve the current left/right preconditioning behavior and true residual reporting.
- Add tests for bad and accepted initial guesses in `testSysGmres.cpp`.
- Update `CHANGELOG.md`.

 
 ## Checklist
 
 
 

     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 ## Further comments

During CUDA validation, the warning build surfaced existing CUDA deprecation warnings in unrelated direct-solver/cuSPARSE/cuSolverRF paths. The modified files compiled without new warnings.

One manual CUDA `sysGmres.exe` run using the `ACTIVSg200` matrix hit an existing Gram-Schmidt zero-norm assertion. The same example passed on CUDA with the smaller `A_asymmetric01/b_asymmetric01` test pair, and the targeted CUDA GMRES tests passed.

### DIFF
--- a/resolve/LinSolverIterativeFGMRES.cpp
+++ b/resolve/LinSolverIterativeFGMRES.cpp
@@ -153,13 +153,25 @@ namespace ReSolve
     int k1         = 0;
 
     real_type   t             = 0.0;
-    real_type   res_norm      = 0.0; // Residual norm used used for convergence
+    real_type   res_norm      = 0.0; // Residual norm used for convergence
     real_type   rhs_norm      = 0.0; // Right-hand side norm used for convergence
     real_type   true_res_norm = 0.0; // True (unpreconditioned) residual norm ||b - Ax|| for reporting
     real_type   true_rhs_norm = 0.0; // True (unpreconditioned) right-hand side norm ||b|| for reporting
+    real_type   initial_rnorm = 0.0;
+    real_type   final_rnorm   = 0.0;
+    real_type   xnorm         = 0.0;
     real_type   tolrel;
     vector_type vec_v(n_);
     vector_type vec_z(n_);
+    vector_type x_initial(x->getSize());
+
+    x_initial.allocate(memspace_);
+    x_initial.copyFromExternal(x, memspace_, memspace_);
+
+    xnorm = vector_handler_->dot(&x_initial, &x_initial, memspace_);
+    xnorm = std::sqrt(xnorm);
+
+    bool check_initial_guess = (xnorm > MACHINE_EPSILON);
 
     // Compute initial residual norm.
     // V[0] = ||b - A*x0||         for right preconditioning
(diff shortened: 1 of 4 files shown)


pelesh (dismissed):
Mark as ready to merge when the work on this feature is complete.

#417 needs to be reviewed and merged, first. Then this branch needs to be rebased.

resolve/LinSolverIterativeFGMRES.cpp
```diff
@@ -153,13 +153,25 @@ namespace ReSolve
...
+    real_type   xnorm         = 0.0;
     real_type   tolrel;
     vector_type vec_v(n_);
     vector_type vec_z(n_);
+    vector_type x_initial(x->getSize());
+
+    x_initial.allocate(memspace_);
+    x_initial.copyFromExternal(x, memspace_, memspace_);
+
+    xnorm = vector_handler_->dot(&x_initial, &x_initial, memspace_);
```
shakedregev: x_norm

resolve/LinSolverIterativeFGMRES.cpp
```diff
@@ -427,6 +462,16 @@ namespace ReSolve
 
       if (!outer_flag)
       {
+        final_rnorm = true_res_norm;
+
+        if (check_initial_guess && final_rnorm > initial_rnorm)
+        {
+          out::warning() << "Iterative solver did not improve the initial guess. Returning the initial guess.\n";
+          x->copyFromExternal(&x_initial, memspace_, memspace_);
+          final_rnorm   = initial_rnorm;
```
shakedregev: r_norm

shakedregev (commented):
Looks good. Make the small corrections and I'll test it.

resolve/LinSolverIterativeFGMRES.cpp
tamar-dewilde: I updated `xnorm` to `x_norm`. I also noticed the existing variable `tolrel` nearby. Do you want that renamed to `tol_rel` as part of this cleanup, or should I leave it unchanged since it was already in the file?

shakedregev: yes

shakedregev (approved):
Great work, all tests pass

PR #429, opened 2026-06-03
Add timing capabilities to the examples

### LINKED ISSUE #326: Add timing capabilities to the examples
### Summary

Add the ability to set a flag that will time the solution of each linear system in `gpuRefactor` and `kluRefactor`.
Use the new code to test both examples on HIP and CUDA. 

### Rationale

There is no way to compare cross platform or cross solver currently.

### Description

The matrices are on Frontier at `/ccs/proj/eng151`.
Use the high resolution clock from the chronos library for timing. Do not include reading or writing from files or printing in the timing.

Test on the gridkit problems there for N=125, 250, 500, and 1k. (Note that this is some internal variable to GridKit. The size of the matrix $`n=36N-2`$. $`nnz\approx 5.5n`$)

We need 9 graphs overall. For each problem we need the solution time and residual for each system with all 4 methods (8 total). The last graph will show the scaling of the average (over all systems) runtime on each problem, for each solver+hardware.

### Additional information

@pelesh - feel free to add details.

tamar-dewilde:
## Description
 
This PR adds optional timing output to the refactor examples for the GridKit timing workflow.

When run with `-t` or `--time`, each example prints one machine-readable timing row per linear system:

```text
TIMING,example,backend,ir_enabled,system,time_ms
```

The timed region excludes file input and summary/timing output. For GPU examples, the active GPU backend is synchronized before reading host-side timing so CUDA/HIP timings include completed device work.

This PR also adds helper scripts under `scripts/timimg/` to parse benchmark logs into CSV and generate the requested timing/residual plots.
 
Closes #326
 
@shakedregev 
 

 ## Proposed changes
 
- Add `-t` / `--time` timing output to `gpuRefactor`.
- Add `-t` / `--time` timing output to `kluRefactor`.
- Add timing support to the remaining refactor examples:
  -gluRefactor
  -sysRefactor
- Add GPU synchronization around timed regions in GPU examples.
- Fix the `kluRefactor -i` path so iterative refinement runs with a valid residual check.
- Add benchmark utilities under `scripts/timing/`:
  - `parse_refactor_logs.py`
  - `plot_refactor_results.py`
  - `README.md`

 
 ## Checklist
 
 
 

     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
     - CPU and HIP warning builds completed cleanly. The CUDA warning build completed, but emitted existing CUDA/cuSPARSE/cuSOLVER deprecation warnings unrelated to this PR. 
 
 ## Further comments
 
I kept the benchmark workflow split into three parts:

```text
example timing output -> saved logs -> Python post-processing scripts
```
This keeps the C++ changes limited to optional timing output and avoids adding Frontier specific benchmark running logic to the examples.

I considered making the Python scripts run and name the benchmark logs directly, but that would require encoding build paths, matrix/RHS prefixes, backend setup, and Frontier specific run details. Keeping the scripts as post-processing utilities makes them easier to reuse for both local smoke tests and Frontier runs.

The Python scripts are post-processing utilities only. They do not run the examples or change solver behavior. Generated logs, CSV files, and plots are intentionally not committed.

AI/tooling note: Since we have not formally discussed LLM-assisted development workflows yet, I am disclosing that I used ChatGPT as a drafting and review aid while developing the Python parsing/plotting utilities more quickly. I reviewed and tested the scripts locally before adding them.


tamar-dewilde:
Update after adding the remaining refactor examples and documentation:

I added timing support for the remaining requested refactor examples and updated the documentation. Then I regenerated the requested nine benchmark plots for GridKit `N=125`, `N=250`, `N=500`, and `N=1000`.

Updates included:

- Added timing support for:
  - `gluRefactor`
  - `sysRefactor`
- Renamed the benchmark utility folder from scripts/issue326/ to scripts/timing/.
- Updated the GridKit timing README with:
  - Frontier CPU/KLU log collection
  - Frontier HIP log collection
  - local CUDA log collection
  - Frontier allocation references
  - OLCF data-transfer references
  - expected log naming
  - CSV parsing
  - plot generation

The plotted methods are:

- `CPU klu`
- `CPU klu_ir`
- `CPU sys_refactor`
- `HIP gpu_refactor`
- `HIP gpu_refactor_ir`
- `HIP sys_refactor`
- `HIP sys_refactor_ir`
- `CUDA gpu_refactor`
- `CUDA gpu_refactor_ir`
- `CUDA glu`
- `CUDA sys_refactor`

Benchmark notes:

Benchmark notes:

-HIP, CPU KLU, and CPU/HIP sysRefactor logs were collected on Frontier using the GridKit matrices in `/ccs/proj/eng151`.
-CUDA logs were collected locally using the same GridKit matrices copied from `/ccs/proj/eng151`.
-While regenerating the full GridKit benchmark set, CUDA `sysRefactor -i` reproducibly hit an existing Gram-Schmidt zero-norm assertion during `N=125`, system 18. Since this is in the shared FGMRES/Gram-Schmidt solver path rather than the timing code, I excluded that partial CUDA `sysRefactor -i` path from the final CSV/plots and will treat it as a separate debug item unless you want it handled in this PR.
-The final CSV combines the completed CPU, HIP, and CUDA logs and produces the nine requested plots:
  -solve time per system for each `N`
  -residual per system for each `N`
  -average solve-time scaling over all systems

average_solve_time_scaling
N125_residual
N125_solve_time
N250_residual
N250_solve_time
N500_residual
N500_solve_time
N1000_residual
N1000_solve_time

shakedregev (commented):
This is great work! I know we didn't have it in the issue, but it would be great to have this for the other `Refactor` codes (glu and sys). 
@pelesh - what do you think about having the python scripts here? It's pretty convenient (maybe we can just change the folder name).

scripts/timing/README.md
```diff
@@ -0,0 +1,505 @@
+# GridKit timing utilities
```
nkoukpaizan: I recommend disentangling `GridKit` and `ReSolve` in this README. The timing utilities themselves are in ReSolve. It's being applied to GridKit data, but that does not have to be the case.

tamar-dewilde: Thanks, that makes sense. I had run a correction script on the README after renaming `scripts/issue326/` to `scripts/timing/`, and the replacement was too broad.

I updated it so the scripts are described as ReSolve refactor timing utilities, with GridKit framed as one example benchmark workflow/data set rather than the purpose of the utilities themselves.

shakedregev (approved):
Great work. I tested this on different matrices also and it still worked.

PR #428, opened 2026-06-02
Fix make test_install failure

pelesh:
## Description
 
 Target `make test_install` fails if the default compiler on the installation machine is different from the compiler Re::Solve is built with, for example, if you build Re::Solve with GCC on a Mac.

 ## Proposed changes
 
 Test script should pass compiler installation to the consumer test, together with all the other build info.

 ## Checklist
 
 
- [x] All tests pass (`make test` and `make test_install` per testing [instructions](https://github.com/ORNL/ReSolve?tab=readme-ov-file#test-and-deploy)). Code tested on
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
- [x] I have manually run the non-experimental examples and verified that residuals are close to machine precision. (In your build directory run: `./examples/<your_example>.exe -h` to get instructions how to run examples). Code tested on:
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
- [x] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [x] The new code follows Re::Solve style guidelines.
- [x] There are unit tests for the new code.
- [x] The new code is documented.
- [x] The feature branch is rebased with respect to the target branch.
- N/A I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR. If this is a minor PR that is part of a larger fix already included in the file, state so.
 
 ## Further comments

PR #434, opened 2026-06-04
Fix potential failure in `make test_install`

pelesh:
## Description
 
If `make test_install` is called before `make install` call, the test will fail giving misleading feedback that something is wrong with the installation configuration. See e.g. #428 

 ## Proposed changes
 
 Enforce installation as a part of the target `test_install`. One-line fix suggested.

 ## Checklist
 

- [ ] All tests pass (`make test` and `make test_install` per testing [instructions](https://github.com/ORNL/ReSolve?tab=readme-ov-file#test-and-deploy)). Code tested on
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
- [ ] I have manually run the non-experimental examples and verified that residuals are close to machine precision. (In your build directory run: `./examples/<your_example>.exe -h` to get instructions how to run examples). Code tested on:
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
- N/A There are unit tests for the new code.
- N/A The new code is documented.
 
 ## Further comments

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -113,6 +113,7 @@ install(FILES ${PROJECT_SOURCE_DIR}/tests/functionality/TestHelper.hpp
 # directory. 3. CXX compiler (propagated so consumer uses same API as ReSolve).
 add_custom_target(
   test_install
+  COMMAND ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR} --target install
   COMMAND ${CONSUMER_PATH}/test.sh ${CMAKE_INSTALL_PREFIX} ${test_data_dir}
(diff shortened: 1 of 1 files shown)


tamar-dewilde (approved):
I tested this with fresh CPU, CUDA, and HIP builds by running test_install without manually running install first. All three passed with local `CMAKE_INSTALL_PREFIX` values, and the fix looks correct to me.

andrewxu319 (approved):
This works on a clean CPU/CUDA build. Called `make test_install` without `make install`.

shakedregev:
Tested and it works. Nice clean fix.

PR #436, opened 2026-06-09
Require explicit memory management from the user and remove sloppy memory management.

### LINKED ISSUE #295: Strictly enforce memory requirements.

andrewxu319:
## Description
Remove automatic memory allocation and memory syncing in various memory management functions and require the user to explicitly manage their own memory.
Resolves #295. 

 
 
 
 
  
 

 ## Proposed changes
Major changes:
- Assert that data is updated in `Vector::getData()` and remove automatic synchronization
- Assert that data is allocated in `copyFromExternal()` and remove automatic allocation for vectors and matrices
- Assert that data is not already allocated in `allocate()` and `allocateMatrixData()` for vectors and matrices and remove auto deallocation
- Add `isUpdated()`, `isAllocated()`, `allocateAll()` to Vector
- Add `allocateAll()` to Sparse
- Update all parts of the code that call these functions and are affected by these changes such that they work correctly

Minor changes:
- Update `.gitignore` to ignore test outputs
- Remove `HyKKTSolver::readMatrixFiles()` which is unnecessary
 
 
 ## Checklist
 
 
 

     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
 
 ## Further comments
 I also tried to add status returns to a few functions in the `VectorHandler` and `LinAlgWorkspace` series of classes, but that crashes the HIP implementation for seemingly unknown reasons.


resolve/hykkt/HyKKTSolver.cpp
```diff
@@ -434,7 +390,7 @@ namespace ReSolve
     spgemm_hgamma_ = new SpGEMM(memspace_, gamma_, ONE);
     spgemm_hgamma_->loadProductMatrices(J_tr_, J_);
     spgemm_hgamma_->loadSumMatrix(H_tilde_);
-    spgemm_hgamma_->loadResultMatrix(&H_gamma_); // H_gamma_ will be created by SpGEMM at this step
+    spgemm_hgamma_->loadResultMatrix(&H_gamma_); // H_gamma_ will be created by SpGEMM atcompute()
```
shakedregev: I don't understand the end of this comment
```suggestion
    spgemm_hgamma_->loadResultMatrix(&H_gamma_); // H_gamma_ will be created by SpGEMM
```

resolve/LinSolverDirectCuSolverRf.cpp
```diff
@@ -234,7 +234,7 @@ namespace ReSolve
    * @return 0 if successful, 1 otherwise
    */
   int LinSolverDirectCuSolverRf::solve(vector_type* rhs, vector_type* x)
-  {
+  { // CAREFUL: ANYTHING THAT CALLS THIS. DELETE THIS COMMENT
```
shakedregev: Do we need this comment or not?

resolve/hykkt/HyKKTSolver.cpp
andrewxu319: `H_gamma_` is created and allocated at `spgemm_hgamma_->compute()`, so this is a note to not try to create it somewhere else.

This is rather confusing, and I was going to remove loadResultMatrix() and instead do something like this:
`H_gamma_ = compute()`
And have `compute()` return the newly created matrix. But it seems that `compute()` is designed to be called multiple times with the same sparsity structure and only create the matrix on the first call, so this wouldn't work.

resolve/LinSolverDirectCuSolverRf.cpp
andrewxu319: Oops, this should be deleted

tests/functionality/testRefactor.cpp
```diff
@@ -180,6 +185,7 @@ int runTest(int argc, char* argv[], std::string& solver_name)
   error_sum += status;
 
   // Solve system (on device where available)
```
shakedregev: This comment seems weird. I think this test is device only.

tests/unit/matrix/MatrixHandlerTests.hpp
```diff
@@ -132,14 +132,23 @@ namespace ReSolve
         testname += matrix_size.str();
 
         matrix::Csr* At = new matrix::Csr(m, n, 2 * std::min(n, m));
-        matrix::Csr* A  = nullptr; // Declare A outside
+        At->allocateAll(memspace_);
+        matrix::Csr* A = nullptr; // Declare A outside
 
         for (real_type val = 0.0; val <= 1.0; val += 1.0)
         { // Use a step to prevent infinite loop
```
shakedregev: Not really part of this PR, but it's not clear to me why we have a loop at all here. It's simply the first thing happening, then the second.

resolve/matrix/Coo.cpp
```diff
@@ -221,37 +221,16 @@ namespace ReSolve
...
-        this->h_row_data_          = new index_type[nnz_current];
-        this->h_col_data_          = new index_type[nnz_current];
-        owns_cpu_sparsity_pattern_ = true;
-      }
-      if (h_val_data_ == nullptr)
+      if ((h_row_data_ == nullptr) || (h_col_data_ == nullptr) || (h_val_data_ == nullptr))
       {
-        this->h_val_data_ = new real_type[nnz_current];
-        owns_cpu_values_  = true;
+        out::error() << "Trying to copy from external COO matrix, but destination (host) is not allocated!\n";
```
superwhiskers, line 226: why aren't we bailing from the function here (and in subsequent instances)? from what i can see, all of the `out::error()` writes this pull request adds don't follow up with a return from the function to avoid entering an invalid state. this should be resolved, unless i'm misunderstanding something.

shakedregev (approved):
Very good work! Tests pass on all platforms. See:
1. Some minor documentation comments. 
2. Some very weird code that was not part of this PR. Should be fixed now or later.

resolve/vector/Vector.cpp
```diff
@@ -290,80 +346,21 @@ namespace ReSolve
     }
 
     /**
-     * @brief get a pointer to HOST or DEVICE vector data.
+     * @brief get a non-const (writable) pointer to HOST or DEVICE vector data.
```
superwhiskers: somewhat confused by this clarification. i think the type signature makes this apparent, but this is a bit of a nitpick

resolve/hykkt/HyKKTSolver.cpp
shakedregev: How about "when calling `->compute()`"

resolve/workspace/LinAlgWorkspaceCUDA.cpp
```diff
@@ -90,10 +92,7 @@ namespace ReSolve
 
   void LinAlgWorkspaceCUDA::setTransposeBufferWorkspace(size_t bufferSize)
   {
-    if (transpose_workspace_ready_)
-    {
-      mem_.deleteOnDevice(transpose_workspace_);
-    }
+    assert(!transpose_workspace_ready_ && "Transpose workspace already set!\n");
```
superwhiskers: this will go away if `NDEBUG` is set, so if this is a load-bearing assertion, it may make sense to not use asssert.

tests/unit/matrix/MatrixHandlerTests.hpp
shakedregev: @pelesh

superwhiskers (commented):
some minor comments.

i ran the cpu-only tests on my local machine, all cpu tests passed. one warning was not addressed:
```
ReSolve/examples/sysRefactor.cpp:33:43: warning: unused parameter ‘hw_backend’ [-Wunused-parameter]
   33 | static void syncDevice(const std::string& hw_backend)
      |                        ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
```

other than that, it seems fine. this warning may be an artifact of a cpu-only build. the gpu build tests are taking a little longer, so i'll leave this review without them.

tests/unit/matrix/MatrixHandlerTests.hpp
andrewxu319: Fixed in the latest commit

andrewxu319:
> some minor comments.
> 
> i ran the cpu-only tests on my local machine, all cpu tests passed. one warning was not addressed:
> 
> ```
> ReSolve/examples/sysRefactor.cpp:33:43: warning: unused parameter ‘hw_backend’ [-Wunused-parameter]
>    33 | static void syncDevice(const std::string& hw_backend)
>       |                        ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
> ```
> 
> other than that, it seems fine. this warning may be an artifact of a cpu-only build. the gpu build tests are taking a little longer, so i'll leave this review without them.

The full function is
```
static void syncDevice(const std::string& hw_backend)
{
#ifdef RESOLVE_USE_CUDA
  if (hw_backend == "CUDA")
  {
    cudaDeviceSynchronize();
  }
#endif

#ifdef RESOLVE_USE_HIP
  if (hw_backend == "HIP")
  {
    hipDeviceSynchronize();
  }
#endif
}
```

So the variable is used if either CUDA or HIP is enabled.

@shakedregev Would it be a good idea to add a `[[maybe_unused]]`? Or just leave it be?

resolve/workspace/LinAlgWorkspaceCUDA.cpp
andrewxu319: Didn't know this, fixed. Thanks.

resolve/matrix/Coo.cpp
superwhiskers: @andrewxu319 not all instances of this have been fixed yet

shakedregev:
> > some minor comments.
> > i ran the cpu-only tests on my local machine, all cpu tests passed. one warning was not addressed:
> > ```
> > ReSolve/examples/sysRefactor.cpp:33:43: warning: unused parameter ‘hw_backend’ [-Wunused-parameter]
> >    33 | static void syncDevice(const std::string& hw_backend)
> >       |                        ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > other than that, it seems fine. this warning may be an artifact of a cpu-only build. the gpu build tests are taking a little longer, so i'll leave this review without them.
> 
> The full function is
> 
> ```
> static void syncDevice(const std::string& hw_backend)
> {
> #ifdef RESOLVE_USE_CUDA
>   if (hw_backend == "CUDA")
>   {
>     cudaDeviceSynchronize();
>   }
> #endif
> 
> #ifdef RESOLVE_USE_HIP
>   if (hw_backend == "HIP")
>   {
>     hipDeviceSynchronize();
>   }
> #endif
> }
> ```
> 
> So the variable is used if either CUDA or HIP is enabled.
> 
> @shakedregev Would it be a good idea to add a `[[maybe_unused]]`? Or just leave it be?

It makes more sense to me to call the function only if you need it. HIP/CUDA `hw_backend`, call this. CPU don't. Anything else should probably throw an error.

superwhiskers:
managed to complete the rocm-enabled tests on my local machine after instituting a workaround for a rocm bug related to my specific amd gpu. builds and runs fine. no leaks outside those present within rocm, runs fine.

there is one concerning thing, and that is the flakiness of two tests on my machine:
- `sys_refactor_hip_asym_test`
- `sys_refactor_hip_test`

primarily the former. this most likely is not related to this pull request, though. just noting this for future reference. it's probably a tolerance thing, unlikely to be anything significant.

examples/sysRefactor.cpp
```diff
@@ -315,67 +317,71 @@ int sysRefactor(int argc, char* argv[])
 
     double solve_time_ms = 0.0;
 
-    syncDevice(hw_backend);
-    auto solve_start = std::chrono::high_resolution_clock::now();
-
-    // Now call direct solver
-    if (i == 1)
+    if (hw_backend == "CUDA" || hw_backend == "HIP")
+      syncDevice(hw_backend);
```
superwhiskers: pretty sure you added a syntax error here

examples/sysRefactor.cpp
```diff
@@ -315,67 +317,71 @@ int sysRefactor(int argc, char* argv[])
...
-    }
-
-    status = solver.solve(vec_rhs, vec_x);
-    syncDevice(hw_backend);
-    auto solve_end = std::chrono::high_resolution_clock::now();
-    solve_time_ms  = std::chrono::duration<double, std::milli>(solve_end - solve_start).count();
-    std::cout << "Triangular solve status: " << status << std::endl;
+      status = solver.solve(vec_rhs, vec_x);
+      if (hw_backend == "CUDA" || hw_backend == "HIP")
+        syncDevice(hw_backend);
```
superwhiskers: likewise

resolve/workspace/LinAlgWorkspaceCUDA.cpp
```diff
@@ -88,15 +93,16 @@ namespace ReSolve
...
-  void LinAlgWorkspaceCUDA::setTransposeBufferWorkspace(size_t bufferSize)
+  int LinAlgWorkspaceCUDA::setTransposeBufferWorkspace(size_t bufferSize)
   {
-    if (transpose_workspace_ready_)
+    if (!transpose_workspace_ready_)
     {
-      mem_.deleteOnDevice(transpose_workspace_);
+      out::error() << "Transpose workspace already set!\n";
+      return 1;
     }
```
superwhiskers, line 102: in this (and the hip version), you've changed the code to read as if the transpose workspace not being ready implies it's already set, and if it is ready, then we need to allocate it.

this looks like an untested change, so i'm somewhat confused what the logic behind this change was. the only call i can find of this method in the codebase is in the `MatrixHanderCuda` code in a path for when the workspace isn't already allocated, which, i assume would trigger this conditional you have here (`!transpose_workspace_ready_`), leading to an error.

i may be wrong on this

superwhiskers (dismissed):

resolve/workspace/LinAlgWorkspaceCUDA.cpp
andrewxu319: Looking at this function
```
  void* LinAlgWorkspaceCUDA::getTransposeBufferWorkspace()
  {
    return transpose_workspace_;
  }
```
And this in the destructor
```
    if (transpose_workspace_ready_)
    {
      mem_.deleteOnDevice(transpose_workspace_);
    }
```
As well as the other times this variable is referenced, it seems that `transpose_workspace_ready_` is indeed just whether the workspace has been allocated. The variable name is misleading, so I'll rename it.

andrewxu319: Although the `!` shouldn't be there.

examples/sysRefactor.cpp
andrewxu319: Fixed, and ran all tests again.

superwhiskers (commented):
still getting this warning on a cpu-only build.
```
ReSolve/examples/sysRefactor.cpp:33:43: warning: unused parameter ‘hw_backend’ [-Wunused-parameter]
   33 | static void syncDevice(const std::string& hw_backend)
      |                        ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
```

everything else looks fine.

shakedregev (approved):
All tests pass and concerns addressed.

PR #441, opened 2026-07-11
Fix memory issues and add config getters in SystemSolver

### LINKED ISSUE #440: Bug: If SystemSolver::preconditionerSetup is called without a preconditioner or iterative solver setup, a segmentation fault occurs

Example:

```c++
ReSolve::LinAlgWorkspaceCpu    linear_workspace;
ReSolve::SystemSolver          lin_solver(&linear_workspace, "klu", "klu", "klu");

lin_solver.initialize();


<...>

lin_solver_.setMatrix( <some matrix> );
lin_solver_.analyze();
lin_solver_.preconditionerSetup("right");
```

Two error messages are printed and a status code variable is incremented, but the function doesn't return with the error - instead it continues to access memory it knows is invalid.

### Re::Solve version

a93fb6571e3542b2fa779983fef0186a034c53c9

Operating system: Ubuntu 24.04.4 LTS
Compiler: g++ (GCC) 13.3.0
Backends: N/A

### Additional information

_No response_

pelesh:
## Description
 
Fixes several memory-safety bugs in `SystemSolver` and completes its solver-configuration query API.

@tamar-dewilde @shakedregev 

I believe this fixes #440 

 ## Proposed changes
 
### Bug fixes

- **`setMatrix()`**: fixed a memory leak — `resVector_` was being overwritten with a new allocation on every call without freeing the previous one.
- **`preconditionerSetup()`**: fixed a null-pointer dereference — the function checked `preconditioner_`/`iterativeSolver_` for null and recorded an error, but then fell through and dereferenced them anyway. Now returns early with the accumulated error status before touching either pointer.
- **`resetPreconditioner()`**: same class of bug — now returns immediately when `preconditioner_` is null instead of calling `preconditioner_->reset(A)` on it.
- **`setSolveMethod()`** and **`setRefinementMethod()`**: fixed dangling-pointer bugs — `iterativeSolver_`/`gs_` were `delete`d without being reset to `nullptr`. Several early-return paths (invalid method, IR unsupported on CPU) left them dangling, which could later cause a double-free in the destructor or a use-after-free in `solve()`. Both pointers are now nulled immediately after `delete`, and `setRefinementMethod()` also unconditionally resets `irMethod_` to `"none"` up front so it can't stay stale and point at a freed solver.

### New functionality

Implemented the previously-declared-but-missing getters: `getRefactorizationMethod()`, `getSolveMethod()`, `getRefinementMethod()`, `getOrthogonalizationMethod()` — mirroring the existing `getFactorizationMethod()`.

These getter help user query the configuration of the `SystemSolver`.
 
 ## Checklist
 
 
 

     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
- N/A (see #444) I have manually run the non-experimental examples and verified that residuals are close to machine precision. (In your build directory run: `./examples/<your_example>.exe -h` to get instructions how to run examples). Code tested on:
     - [ ] CPU backend
     - [ ] CUDA backend
     - [ ] HIP backend
 
 ## Further comments
 
- This PR does **not** address code style issues in `SystemSolver` so they don't distract from the bug fix review.
- This is mainly fixing deficiency in `SystemSolver` class, so no entries in CHANGELOG.

### DIFF
--- a/resolve/SystemSolver.cpp
+++ b/resolve/SystemSolver.cpp
@@ -63,6 +63,14 @@ namespace ReSolve
       precondition_method_   = "none";
     }
 
+    if ((solve == "randgmres" || solve == "fgmres") && (ir != "none"))
+    {
+      out::warning() << "Incorrect input: "
+                     << "Iterative refinement cannot be enabled together with an iterative solve method.\n"
+                     << "Setting refinement method to 'none' ...\n";
+      irMethod_ = "none";
+    }
+
     // Instantiate handlers
     matrixHandler_ = new MatrixHandler(workspaceCpu_);
     vectorHandler_ = new VectorHandler(workspaceCpu_);
@@ -93,6 +101,14 @@ namespace ReSolve
                      << "Setting both to 'none' ...\n";
       refactorizationMethod_ = "none";
       precondition_method_   = "none";
+    }
+
+    if ((solve == "randgmres" || solve == "fgmres") && (ir != "none"))
+    {
+      out::warning() << "Incorrect input: "
+                     << "Iterative refinement cannot be enabled together with an iterative solve method.\n"
+                     << "Setting refinement method to 'none' ...\n";
+      irMethod_ = "none";
     }
 
     // Instantiate handlers
@@ -126,6 +142,14 @@ namespace ReSolve
                      << "Setting both to 'none' ...\n";
       refactorizationMethod_ = "none";
       precondition_method_   = "none";
+    }
+
+    if ((solve == "randgmres" || solve == "fgmres") && (ir != "none"))
+    {
+      out::warning() << "Incorrect input: "
+                     << "Iterative refinement cannot be enabled together with an iterative solve method.\n"
+                     << "Setting refinement method to 'none' ...\n";
+      irMethod_ = "none";
     }
 
     // Instantiate handlers
@@ -804,6 +828,14 @@ namespace ReSolve
 
     if (method == "none")
       return;
+
+    if (solveMethod_ == "randgmres" || solveMethod_ == "fgmres")
+    {
+      out::warning() << "Iterative refinement cannot be enabled together with an "
+                     << "iterative solve method ('randgmres' or 'fgmres'). "
+                     << "Keeping refinement method 'none'.\n";
+      return;
+    }
 
     if (memspace_ == "cpu")
     {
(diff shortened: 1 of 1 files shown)


resolve/SystemSolver.cpp
```diff
@@ -181,6 +181,7 @@ namespace ReSolve
   {
     int status = 0;
     A_         = A;
+    delete resVector_;
     resVector_ = new vector_type(A->getNumRows());
```
alexander-novo: As far as I can tell, this is the only place where `resVector_` is set, so the first time this function is called, it will try to `delete` a `nullptr`.

Similarly, I think the `delete resVector_;` in the destructor should be protected as well (in case `setMatrix` was never called).

alexander-novo (changes requested):

resolve/SystemSolver.cpp
pelesh: Addressed in https://github.com/ORNL/ReSolve/pull/441/commits/78b95fbc18573d41b3ca61703ebec4ec724dcdf6

JeffZ594: Is a similar guard also necessary at the destructor?

pelesh: Fixed that, too, thanks!

I think a better overall solution would be to have a vector `resVector_` and not a pointer as the member. Then in `setupMatrix` method we would use `resize` instead of deleting and re-instantiating `resVector_` object. I'll take a look at that in a separate PR.

resolve/SystemSolver.cpp
```diff
@@ -63,6 +63,14 @@ namespace ReSolve
       precondition_method_   = "none";
     }
 
+    if ((solve == "randgmres" || solve == "fgmres") && (ir != "none"))
+    {
+      out::warning() << "Incorrect input: "
+                     << "Iterative refinement cannot be enabled together with an iterative solve method.\n"
+                     << "Setting refinement method to 'none' ...\n";
+      irMethod_ = "none";
+    }
```
shakedregev, line 72: I think I prefer an error here. Anyone who calls iterative refinement with an iterative solver likely made a mistake. This implementation decides the mistake is on the refinement, but it could be on the solver.

alexander-novo (approved):

resolve/SystemSolver.cpp
alexander-novo: I agree

pelesh: This is following current approach in `SystemSolver`. We will eventually rework this interface altogether so that user cannot enter conflicting inputs to begin with.

PR #442, opened 2026-07-11
Protect MemoryHandler alias with a namespace

pelesh:
## Description
 
Protect `MemoryHandler` alias with `ReSolve` namespace. 
 

 ## Proposed changes
 
 ` MemoryHandler` was placed in global namespace by accident. Now it is within `ReSolve` namespace.

 ## Checklist
 
 
 

     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
- N/A (see #444) I have manually run the non-experimental examples and verified that residuals are close to machine precision. (In your build directory run: `./examples/<your_example>.exe -h` to get instructions how to run examples). Code tested on:
     - [ ] CPU backend
     - [ ] CUDA backend
     - [ ] HIP backend
 
 ## Further comments

### DIFF
--- a/resolve/MemoryUtils.hpp
+++ b/resolve/MemoryUtils.hpp
@@ -131,10 +131,8 @@ namespace ReSolve
 // Check if GPU support is enabled in Re::Solve and set appropriate device memory manager.
 #if defined RESOLVE_USE_CUDA
 #include <resolve/cuda/CudaMemory.hpp>
-using MemoryHandler = ReSolve::MemoryUtils<ReSolve::memory::Cuda>;
 #elif defined RESOLVE_USE_HIP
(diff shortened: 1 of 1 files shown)


alexander-novo:
Does `MemoryHandler` not get consumed by anything in the code base? Is it just for users' sake?

pelesh:
> Does `MemoryHandler` not get consumed by anything in the code base? Is it just for users' sake?

It is used in workspaces, matrix/vector classes and handlers, and everywhere else where memory management, including host-device transfers is required. It is not supposed to be exposed to the users.

alexander-novo:
Ah okay but all this stuff is already in the correct namespace so they didn’t need to be changed. Makes sense.

alexander-novo (approved):

PR #459, opened 2026-07-21
cuDSS implementations of Cholesky solver and refactorization

andrewxu319:
## Description
 
Added cuDSS implementations of Cholesky solver and refactorization on CUDA. These are currently implemented in cuSolver, which is an older library. Updated unit tests for refactorization to test both the cuDSS and cuSolver implementations.
 
 
 
 
  
 

 ## Proposed changes
Cholesky solver: rewrote with cuDSS, but retained cuSolver implementation. The user can pass the `use_cudss` parameter in the constructor to decide which version to use. It is true by default. The cuDSS implementation is around 4x the speed of the cuSolver implementation.

Refactorization: added `LinSolverDirectCuDssRf` class as an independent solver. It is very slightly slower than the cuSolver implementation. The interface is kept unchanged, but the following variables are actually unused:
- Matrices `L` and `U` are allocated internally by cuDSS
- Permutation vector `Q` is unused, only `P`. cuDSS only takes one permutation vector and computes the other
- `nboost` / `pivot_boost` have no equivalents in cuDSS
Other changes:
- Updated `SystemSolver.cpp` to accept `cudssrf` as a factorization method
- Updated `testRefactor.cpp` and `testSysRefactor.cpp` to run tests on both implementations

 
 
 ## Checklist
 
 
 

     - [N/A] CPU backend
     - [x] CUDA backend
     - [N/A] HIP backend
     - [N/A] CPU backend
     - [x] CUDA backend
     - [N/A] HIP backend
 
 ## Further comments

### DIFF
--- a/cmake/ReSolveFindCudaLibraries.cmake
+++ b/cmake/ReSolveFindCudaLibraries.cmake
@@ -4,12 +4,15 @@
 add_library(resolve_cuda INTERFACE)
 
 find_package(CUDAToolkit REQUIRED)
-find_package(cudss REQUIRED)
 
 target_link_libraries(
   resolve_cuda INTERFACE CUDA::cusolver CUDA::cublas CUDA::cusparse
-                         CUDA::cudart cudss
+                         CUDA::cudart
 )
+
+if(RESOLVE_USE_CUDSS)
+  target_link_libraries(resolve_cuda INTERFACE cudss)
+endif()
 
 if(RESOLVE_USE_PROFILING)
   target_link_libraries(resolve_cuda INTERFACE CUDA::nvToolsExt)
--- a/cmake/ReSolveFindCudaLibraries.cmake
+++ b/cmake/ReSolveFindCudaLibraries.cmake
@@ -4,10 +4,11 @@
 add_library(resolve_cuda INTERFACE)
 
 find_package(CUDAToolkit REQUIRED)
+find_package(cudss REQUIRED)
 
 target_link_libraries(
   resolve_cuda INTERFACE CUDA::cusolver CUDA::cublas CUDA::cusparse
-                         CUDA::cudart
+                         CUDA::cudart cudss
 )
 
 if(RESOLVE_USE_PROFILING)
--- a/examples/cudssRefactor.cpp
+++ b/examples/cudssRefactor.cpp
@@ -1,5 +1,6 @@
 #include <chrono>
 #include <cmath>
+#include <cuda_runtime.h>
 #include <iomanip>
 #include <iostream>
 #include <sstream>
@@ -16,8 +17,6 @@
 #include <resolve/vector/Vector.hpp>
 #include <resolve/vector/VectorHandler.hpp>
 #include <resolve/workspace/LinAlgWorkspace.hpp>
-
-#include <cuda_runtime.h>
 
 /// Prints help message describing system usage.
 void printHelpInfo()
@@ -160,11 +159,11 @@ int cudssRefactor(int argc, char* argv[])
 
   // Create system solver
   ReSolve::SystemSolver solver(&workspace,
-                               "klu",    // factorization
+                               "klu",     // factorization
                                "cudssrf", // refactorization
                                "cudssrf", // triangular solve
-                               "none",   // preconditioner (always 'none' here)
-                               "none");  // iterative refinement
+                               "none",    // preconditioner (always 'none' here)
+                               "none");   // iterative refinement
 
   if (is_iterative_refinement)
   {
--- a/examples/cudssRefactor.cpp
+++ b/examples/cudssRefactor.cpp
@@ -145,7 +145,7 @@ int cudssRefactor(int argc, char* argv[])
   workspace_type workspace;
   workspace.initializeHandles();
 
-  // Create a helper object (computing errors, printing summaries, etc.)
+  // Create a helper object (for computing errors, printing summaries, etc.)
   ExampleHelper<workspace_type> helper(workspace);
   std::cout << "cudssRefactor with CUDA backend\n";
 
--- a/examples/cudssRefactor.cpp
+++ b/examples/cudssRefactor.cpp
@@ -1,0 +1,302 @@
+#include <chrono>
+#include <cmath>
+#include <iomanip>
+#include <iostream>
+#include <sstream>
+#include <string>
+
+#include "ExampleHelper.hpp"
+#include <resolve/LinSolverDirectKLU.hpp>
+#include <resolve/Profiling.hpp>
+#include <resolve/SystemSolver.hpp>
+#include <resolve/matrix/Csr.hpp>
+#include <resolve/matrix/MatrixHandler.hpp>
+#include <resolve/matrix/io.hpp>
+#include <resolve/utilities/params/CliOptions.hpp>
+#include <resolve/vector/Vector.hpp>
+#include <resolve/vector/VectorHandler.hpp>
+#include <resolve/workspace/LinAlgWorkspace.hpp>
+
+#include <cuda_runtime.h>
+
+/// Prints help message describing system usage.
+void printHelpInfo()
+{
+  std::cout << "\ncudssRefactor.exe loads from files and solves a series of linear systems.\n\n";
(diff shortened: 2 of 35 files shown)


shakedregev:
Fixed the changelog "conflict". Reviewing and testing this now. Please pull before making new changes.

cmake/ReSolveFindCudaLibraries.cmake
```diff
@@ -4,10 +4,11 @@
 add_library(resolve_cuda INTERFACE)
 
 find_package(CUDAToolkit REQUIRED)
+find_package(cudss REQUIRED)
```
shakedregev: I think it is better to not require cudss unless the user specifically requests a cudss enabled build.

resolve/LinSolverDirectCuDssRf.cpp
```diff
@@ -0,0 +1,464 @@
...
+  }
+
+  /**
+   * @brief Destructor for LinSolverDirectCuDssRf
+   *
+   * Destroys the cuDssRf handle, config, and data, then deletes the permutation vectors
+   * from the device memory.
+   *
+   * @pre The cuDssRf handle, config, and data have been created.
+   * @post The cuDssRf handle, config, and data have destroyed.
```
shakedregev:
```suggestion
   * @post The cuDssRf handle, config, and data have been destroyed.
```

resolve/LinSolverDirectCuDssRf.cpp
```diff
@@ -0,0 +1,464 @@
...
+   *
+   * @post The matrix A is refactorized.
+   *
+   * @return 0 if successful, 1 otherwise
+   */
+  int LinSolverDirectCuDssRf::refactorize()
+  {
+    assert(A_ != nullptr && "Matrix A is null!");
+    assert(A_->getNumRows() > 0 && "Matrix A must have positive row count!");
+    assert(A_->getNnz() > 0 && "Matrix A must have positive nonzero count!");
```
shakedregev:
```suggestion
    assert(A_->getNumRows() > 0 && "Matrix A must have positive row count!");
    assert(A_->getNnz() > 0 && "Matrix A must have positive number of nonzeros!");
```
`positive nonzero` sounds very confusing

resolve/LinSolverDirectCuDssRf.cpp
```diff
@@ -0,0 +1,464 @@
...
+  int LinSolverDirectCuDssRf::solve(vector_type* rhs, vector_type* x)
+  {
+    index_type n = A_->getNumRows();
+
+    int error_sum = 0;
+
+    cudssMatrix_t rhs_descr;
+    cudssMatrix_t x_descr;
+    error_sum += cudssMatrixCreateDn(&rhs_descr, n, 1, n, rhs->getData(memory::DEVICE), CUDA_R_64F, CUDSS_LAYOUT_COL_MAJOR);
+    error_sum += cudssMatrixCreateDn(&x_descr, n, 1, n, x->getData(memory::DEVICE), CUDA_R_64F, CUDSS_LAYOUT_COL_MAJOR);
```
shakedregev, line 197: Nothing to fix here, but I noted that CUDA doesn't support row major. This could cause conversion issues with the code down the line.

shakedregev (changes requested):
Please make a new example which uses cudss. Ensure that the old examples with cuSolver still work with no changes. Also, unless you have tested and shown that transposing on the host is faster, I would transpose on the device. I expect it to be faster for large matrices. No need to reimplement everything, just call our existing functions.

andrewxu319:
> Please make a new example which uses cudss. Ensure that the old examples with cuSolver still work with no changes. Also, unless you have tested and shown that transposing on the host is faster, I would transpose on the device. I expect it to be faster for large matrices. No need to reimplement everything, just call our existing functions.

What are you referring to by transpose?

andrewxu319:
> Please make a new example which uses cudss. Ensure that the old examples with cuSolver still work with no changes. Also, unless you have tested and shown that transposing on the host is faster, I would transpose on the device. I expect it to be faster for large matrices. No need to reimplement everything, just call our existing functions.

Done. The new example `cudssRefactor` is a modified copy of `sysRefactor`.

resolve/hykkt/cholesky/CholeskySolverCuda.cpp
```diff
@@ -13,87 +13,182 @@ namespace ReSolve
...
+      if (use_cudss_)
+      {
+        cudssDataDestroy(cudss_handle_, cudss_data_);
+        cudssConfigDestroy(cudss_config_);
+        cudssDestroy(cudss_handle_);
+        cudssMatrixDestroy(descr_A_cudss_);
+        cudssMatrixDestroy(descr_b_);
+        cudssMatrixDestroy(descr_x_);
+      }
+      else
```
shakedregev: what is this else for? it seems within the nested if, but then the commands are outside the outer if

andrewxu319: If `RESOLVE_USE_CUDSS` is defined, there is an if/else that checks `use_cudss_` and takes one branch. If `RESOLVE_USE_CUDSS` is not defined, there is no if/else, so the else is compiled out.

resolve/hykkt/cholesky_cudss/CholeskySolverCuDss.cpp
```diff
@@ -0,0 +1,92 @@
+/**
+ * @file CholeskySolverCuDss.cpp
+ * @author Adham Ibrahim (ibrahimas@ornl.gov)
```
shakedregev:
```suggestion
```

resolve/SystemSolver.cpp
```diff
@@ -316,6 +319,12 @@ namespace ReSolve
     else if (refactorizationMethod_ == "cusolverrf")
     {
       refactorizationSolver_ = new ReSolve::LinSolverDirectCuSolverRf();
+#ifdef RESOLVE_USE_CUDSS
+    }
+    else if (refactorizationMethod_ == "cudssrf")
+    {
+      refactorizationSolver_ = new ReSolve::LinSolverDirectCuDssRf();
+#endif
```
shakedregev, line 327: I find this writing style very confusing, though I see you are copying it and see why it works. I really this brackets should be within #ifdef statements.

resolve/SystemSolver.cpp
```diff
@@ -508,6 +521,17 @@ namespace ReSolve
...
+    else if (refactorizationMethod_ == "cudssrf")
+    {
+      LinSolverDirectCuDssRf* Rf = dynamic_cast<LinSolverDirectCuDssRf*>(refactorizationSolver_);
+      Rf->setNumericalProperties(1e-14, 1e-1);
+
+      status += refactorizationSolver_->setup(A_, L_, U_, P_, Q_);
+
+      is_solve_on_device_ = false;
+    }
+#endif
```
shakedregev, line 534: This way looks much better and is easier to follow.

tests/unit/hykkt/runHykktCholeskyCuDssTests.cpp
```diff
@@ -0,0 +1,55 @@
+/**
+ * @file runHykktCholeskyCuDssTests.hpp
+ * @author Shaked Regev (regevs@ornl.gov)
+ * @author Adham Ibrahim (ibrahimas@ornl.gov)
```
shakedregev:
```suggestion
 * @author Andrew Xu (xua1@ornl.gov)
```

shakedregev (approved):
Great work. Checked the tests and examples

examples/cudssRefactor.cpp
```diff
@@ -0,0 +1,301 @@
...
+            << ", total number of matrices: " << num_systems << "\n"
+            << "Family rhs file name: " << rhs_pathname
+            << ", total number of RHSes: " << num_systems << "\n";
+
+  int status = 0;
+
+  workspace_type workspace;
+  workspace.initializeHandles();
+
+  // Create a helper object (computing errors, printing summaries, etc.)
```
shakedregev:
```suggestion
  // Create a helper object (for computing errors, printing summaries, etc.)
```

Issue #387, opened 2025-10-10
Bug: 38/49 tests fail on CUDA when using address sanitizers.


shakedregev:
### Description how to reproduce the bug

        
```
cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DRESOLVE_USE_ASAN=yes -DRESOLVE_USE_UBSAN=yes -DRESOLVE_USE_KLU=yes -DRESOLVE_USE_CUDA=yes -DRESOLVE_USE_LUSOL=yes -DCMAKE_CUDA_ARCHITECTURES=70 -DCMAKE_INSTALL_PREFIX=../bin -DSUITESPARSE_DIR=/usr/lib/x86_64-linux-gnu -DSUITESPARSE_INCLUDE_DIR=/usr/include/suitesparse  ../ReSolve/
make -j
ctest -j
```

Memory leaks such as:
```
Direct leak of 2996 byte(s) in 4 object(s) allocated from:
    #0 0x7fb54b596887 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x7fb515e8b810  (/usr/local/cuda/lib64/libcusparse.so.12+0xa8b810)

Direct leak of 408 byte(s) in 1 object(s) allocated from:
    #0 0x7fb54b596a57 in __interceptor_calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:154
    #1 0x7fb4d318c004  (/lib/x86_64-linux-gnu/libcuda.so.1+0x2d5004)
    #2 0x7fb4d31cb151  (/lib/x86_64-linux-gnu/libcuda.so.1+0x314151)
    #3 0x7fb4d31bc92f in cuInit (/lib/x86_64-linux-gnu/libcuda.so.1+0x30592f)
    #4 0x7fb532afd8c9  (/usr/local/cuda/lib64/libcublas.so.12+0x4fd8c9)
    #5 0x7fb532afda47  (/usr/local/cuda/lib64/libcublas.so.12+0x4fda47)
    #6 0x7fb51460fee7 in __pthread_once_slow nptl/pthread_once.c:116

```

The same errors do not appear when the address sanitizer is not enabled, or alternatively during CPU only builds.

### Re::Solve version

All versions

### System and environment details

Build environment:
  - pdesx system
  - CUDA 12.9
  - gcc 11.4.0

### Additional information

_This seems like NVIDIA's error._

PR #461, opened 2026-07-29
Fix cuSPARSE handle and test buffer leaks under ASAN

zjin-lcf:
## Summary

Addresses #387. When ReSolve is built with the address sanitizer (`-DRESOLVE_USE_ASAN=yes`) on CUDA, many unit tests fail with leaks reported from `libcusparse`. The root cause is that the cuSPARSE ILU0 direct solver never releases the cuSPARSE resources it creates.

`LinSolverDirectCuSparseILU0::setup()` allocates:
- a matrix descriptor (`descr_A_`),
- two sparse-matrix descriptors (`mat_L_`, `mat_U_`),
- two SpSV descriptors (`descr_spsv_L_`, `descr_spsv_U_`),
- a `csrilu02Info_t` info structure (`info_A_`),
- and three device work buffers (`buffer_`, `buffer_L_`, `buffer_U_`),

but the destructor only freed `d_aux1_`, `d_aux2_`, and `d_ILU_vals_`. Every solver instance therefore leaked the descriptors, info struct, and buffers, which ASAN attributes to the cuSPARSE library.

## Changes

- Destroy all cuSPARSE descriptors/info and free the device work buffers in `~LinSolverDirectCuSparseILU0()`.
- Free the scratch host buffer allocated in the device path of the sparse matrix copy test (`tests/unit/matrix/SparseTests.hpp`), which leaked on device builds.

This mirrors the approach taken for the rocSPARSE analog in #388.

## Notes

Some of the allocations reported by ASAN in the issue originate inside NVIDIA libraries (`libcuda`/`libcublas` one-time initialization via `cuInit`) and are outside ReSolve's control; those are not addressed here.

## Test plan

- [ ] Configure with `-DRESOLVE_USE_ASAN=yes -DRESOLVE_USE_UBSAN=yes -DRESOLVE_USE_CUDA=yes` and run `ctest -j`; verify the previously failing tests no longer report leaks from the ILU0 solver or the sparse matrix copy test.

### DIFF
--- a/resolve/LinSolverDirectCuSparseILU0.cpp
+++ b/resolve/LinSolverDirectCuSparseILU0.cpp
@@ -19,6 +19,55 @@ namespace ReSolve
     mem_.deleteOnDevice(d_aux1_);
     mem_.deleteOnDevice(d_aux2_);
     mem_.deleteOnDevice(d_ILU_vals_);
+
+    if (buffer_ != nullptr)
+    {
+      mem_.deleteOnDevice(buffer_);
+      buffer_ = nullptr;
+    }
+    if (buffer_L_ != nullptr)
+    {
+      mem_.deleteOnDevice(buffer_L_);
+      buffer_L_ = nullptr;
+    }
+    if (buffer_U_ != nullptr)
+    {
+      mem_.deleteOnDevice(buffer_U_);
+      buffer_U_ = nullptr;
+    }
+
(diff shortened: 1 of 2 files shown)


pelesh:
@zjin-lcf, thank you for your contribution. Please re-open the pull request when reproducible test results (before and after) are available together with hardware specs and CUDA version. Closing for now.

PR #472, opened 2026-08-13
Fix compiler warnings for 0.99.3 release

tamar-dewilde:
## Description
 
Cleans up compiler warnings found during 0.99.3 release testing.
 
@pelesh 
 

 ## Proposed changes
 
- Fix the CPU unused parameter warning in `sysRefactor`.
- Fix HIP unused parameter and variable warnings.
- Make HIP signed/unsigned conversions explicit.
 
 ## Checklist
 
 
 

     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
     - [x] CPU backend
     - [x] CUDA backend
     - [x] HIP backend
     - [x] CPU: 0 warnings
     - [x] HIP: 0 warnings
     - [ ] CUDA: deprecation warnings remain
   - No CHANGELOG update needed since this is only a warning cleanup for the 0.99.3 release.
 
 ## Further comments
 
CPU and HIP compile cleanly with `-Wall -Wpedantic -Wconversion -Wextra`.

CUDA still reports the expected cuSOLVER/cuSolverRF deprecation warnings recommending cuDSS. There are also deprecation warnings from the cuSPARSE `csrilu02` ILU0 interface. Removing those would require a larger API update, so I left them unchanged for now. Let me know if you would like me to include that cleanup before the release.

### DIFF
--- a/resolve/hip/hipVectorKernels.hip
+++ b/resolve/hip/hipVectorKernels.hip
@@ -20,7 +20,7 @@ namespace ReSolve {
 
       __global__ void set_array_to_const(index_type n, real_type val, real_type* arr)
       {
-        index_type i = blockIdx.x * blockDim.x + threadIdx.x;
+        index_type i = static_cast<index_type>(blockIdx.x * blockDim.x + threadIdx.x);
         while (i < n)
         {
           arr[i] = val;
@@ -40,7 +40,7 @@ namespace ReSolve {
       */
       __global__ void addConst(index_type n, real_type val, real_type* arr)
       {
-        index_type i = blockIdx.x * blockDim.x + threadIdx.x;
+        index_type i = static_cast<index_type>(blockIdx.x * blockDim.x + threadIdx.x);
         if(i < n)
         {
           arr[i] += val;
@@ -61,7 +61,7 @@ namespace ReSolve {
                                       real_type* vec)
       {
         // Get the index of the element to be processed
-        index_type idx = blockIdx.x * blockDim.x + threadIdx.x;
+        index_type idx = static_cast<index_type>(blockIdx.x * blockDim.x + threadIdx.x);
 
         // Check if the index is within bounds
         if (idx < n) {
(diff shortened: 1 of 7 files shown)


resolve/hip/hipVectorKernels.hip
```diff
@@ -180,7 +180,7 @@ namespace ReSolve {
       const int block_size = 256;
       int num_blocks = (n + block_size - 1) / block_size;
       // Launch the kernel
-      kernels::scale<<<num_blocks, block_size>>>(n, diag, vec);
+      kernels::scale<<<static_cast<unsigned int>(num_blocks), static_cast<unsigned int>(block_size)>>>(n, diag, vec);
```
pelesh: I think a better solution here would be to define
```c++
const size_t block_size = 256;
const size_t num_blocks = (n + block_size - 1) / block_size;
```
That would eliminate the need for `static_cast`s.

Similar to the subsequent instances.

resolve/hip/hipVectorKernels.hip
```diff
@@ -152,15 +152,15 @@ namespace ReSolve {
...
       num_blocks = (n + block_size - 1) / block_size;
-      hipLaunchKernelGGL(kernels::set_array_to_const, dim3(num_blocks), dim3(block_size), 0, 0, n, c, v);
+      hipLaunchKernelGGL(kernels::set_array_to_const, dim3(static_cast<unsigned int>(num_blocks)), dim3(static_cast<unsigned int>(block_size)), 0, 0, n, c, v);
     }
 
     void addConst(index_type n, real_type val, real_type* arr)
     {
       index_type num_blocks;
       index_type block_size = 512;
       num_blocks = (n + block_size - 1) / block_size;
```
pelesh: I would define `num_blocks` and `block_size` as `const size_t` type. That would eliminate the need for static casts.

pelesh (approved):
Great job, thanks!

I think there is a better way to do conversions from Re::Solves `index_type` to native HIP/CUDA index types but this is beyond the scope of this PR.

I left suggestions how type casts could be done more elegantly but haven't. test them myself. You may want to check that quickly.

shakedregev (approved):
Tested, this works.

tamar-dewilde:
Applied the review suggestions and retested. Everything still passes.

